PR #17, opened 2022-05-02
Add a note about memory address for other hisilicon boards.

dkuku:
Add a note about memory address for other boards.


ystinia:
Thank you !

PR #23, opened 2022-07-03
Reformat markdown to fit 80 chars limit

themactep:
We should keep markdown text narrow, as if typewrited on a list of paper, so it is suitable for reading on text-only terminals and for printing without rendering md files into html. And it's easier to make changes later from a tablet.


ZigFisher:
Thank you !

PR #41, opened 2022-07-21
Add overview section to install page, and expand.

linuxtim:
Make it clear that there are now multiple methods of firmware installation, and provide an overview of them before giving detailed instructions for the U Boot + UART + TFTP method.  Tweak title for install page from frontpage.


themactep:
Thank you, Tim. This addition makes perfect sense. Much appreciated.

PR #75, opened 2023-03-14
Alternative way to get root access

naksper:
if we do not have a root pass, it is possible to enter the full working system, with a custom pass for research purposes


FlyRouter:
Thank you !

FlyRouter (approved):

PR #121, opened 2023-08-18
Update gpio-settings.md

naksper:
added gpio setting for Hi3518ev100


naksper:
added gpio settings info for Hi3518ev100

FlyRouter (approved):
Thank you

PR #151, opened 2023-09-15
Update fpv.md

MarioFPVdev:
OpenHD dropped support for Jetson Nano.
Added a few grammar corrections.

### DIFF
--- a/en/fpv.md
+++ b/en/fpv.md
@@ -22,27 +22,27 @@ The receiver in this mode operates in the so-called `monitor mode`, where it rec
 
 ## Revolution from OpenIPC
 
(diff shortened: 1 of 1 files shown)


FlyRouter (commented):
Maybe we should leave out the mention of "and a Jetson Nano" and then the Jetson lovers will come and start all over again ? Why not leave both types of comparisons and references ? Otherwise, I agree with the corrections. Thank you.

FlyRouter:
> OpenHD dropped support for Jetson Nano. Added a few grammar corrections.

Ah, I read this post. Performing the merge.

PR #162, opened 2023-09-21
HomeAssistant View & PTZ command

lucasdellasantina:
Add how to install camera to home assistant and How to move the camera from a visual interface in home, assist through an ssh command


themactep:
Thank you!

PR #175, opened 2023-10-07
Fix minor error, "install-deps" -> "deps"

tenten8401:
The makefile only has "deps" in it, not "install-deps"
image


themactep:
thank you.

PR #191, opened 2023-10-13
Syntax correction and menu item

HKR1987:
Syntax correction in fpv-gs-ubuntu.md and add menu item in readme.md


HKR1987:
already done :)

PR #197, opened 2023-10-20
Add IRSTATUS pin for Hi3518Ev200 XM board

max-ermilov:
On XM IPG-50HV20PET-S (probably all IPG-50HV20PEx-x) GPIO24 corresponds to IRSTATUS input


FlyRouter (approved):
Thank you

PR #199, opened 2023-10-23
Update sigmastar-unbrick.md

mixatronik:


viktorxda:
You don't the CH341A, this is built into snander itself now.

viktorxda (changes requested):
You don't need the CH341A kernel modules anymore, this is built into snander itself now.

FlyRouter:
Only @viktorxda decides on approval or need for corrections

viktorxda:
Like previously mentioned, this is no longer needed. snander itself can now access the CH341A I2C.

mixatronik:
But without modules there is no i2c device. Or am I doing something wrong?

viktorxda:
See here: [snander-mstar/ch341a_i2c.c](https://github.com/viktorxda/snander-mstar/blob/master/src/ch341a_i2c.c)
This was previously being handled by the external CH341A modules.

PR #251, opened 2023-11-10
Update Tapo C310v2

nikdoof:
Update of the Tapo C310v2 data based on the example I purchased and flashed with OpenIPC.


themactep:
Thank you!

PR #288, opened 2023-12-27
Auto night mode

thegroove:
Article for setting up automatic night mode switching on devices without a light sensor


FlyRouter (approved):
Thank you

PR #291, opened 2023-12-31
Create step-by-step FPV setup guide

CurryKitten:
As per title - I've written a document design to give new users an easier time when attempting to setup the necessary hardware/software for the first time.

Despite my best efforts to squash these commits into a single one, it seems unwilling to do so!


FlyRouter (approved):
Thank you very much, integrated

PR #297, opened 2024-01-08
Correct error in USB wiring description

CurryKitten:
I'd written the USB connections incorrectly, so people may have been trying to cross d-/d+ instead of joining them.  Corrected this


themactep:
Thank you.

PR #300, opened 2024-01-15
Sensor driver installation example

RoboSchmied:
This example shows how to install image sensor drivers.
Sc223a will be installed onto a gk7205v210 board.


FlyRouter (approved):
Thank you

PR #305, opened 2024-01-19
Update building.md

PetruSoroaga:
Add info on how to flash/install the build image.


themactep:
`firstboot` is `sysupgrade -n` so it is either one or another.

FlyRouter (approved):
Thanks, patch accepted.
But as Paul said, perhaps you need to write in a little more detail which paragraph the mention of the -n key refers to
We often deal with untrained users who will be hesitant if the text is not intended for babies.. ;)

PR #306, opened 2024-01-20
add info + change fps

RoboSchmied:
add info where to find fps value
change 25fps to new default value (30fps) for sc223a


FlyRouter (approved):
Thank you

PR #338, opened 2024-03-09
Updated instruction for Smartwares CIP-37210

cat101:
1. Updated instructions to work with the newer releases of OpenIPC
2. Added GPIO settings to make it simpler for beginners


FlyRouter (approved):
Thank you

PR #343, opened 2024-03-19
Add some info about XM IPG-G4-WR-BL

Pun0110:
Add some info about XM IPG-G4-WR-BL.
First PR, so maybe bugs here... :-/


FlyRouter (approved):
Thank you

PR #357, opened 2024-04-19
Update network-settings.md

Ocean572:
Added custom network configuration file for gk7205v300 + atbm6032


FlyRouter (approved):
Thank you very much for your addition, however please note that we have switched to using profiles, which you can find in the /etc/wireless/usb directory and you should make one for your device if it does not already exist.
Thank you.

PR #358, opened 2024-04-20
Add my work on audio

snokvist:
Please review some notes on audio on ssc338q-imx415. Unfortunately the GST commands are developed for PC (x86) / steamdeck but should be possible to adapt.


FlyRouter (approved):
Thank you

PR #366, opened 2024-05-05
Update fpv-audio.md

snokvist:
Short and simple, it just works now. Strange behavior if not adapting to the two-payload stream when audio is enabled though but will be up to GS developers to handle.


FlyRouter (approved):
Thank you, sir

PR #367, opened 2024-05-13
Remove SIGUSR1 mention from majestic-streamer.md

Vavooon:
As @widgetii said, SIGUSR1 is no more supported by Majestic.
Entire process hangs if used, and the device restarts if watchdog is enabled.


FlyRouter (commented):
Dear majestic developers and integration consultants.
Please check this statement to see if it is true.
And also, when you have time, try to make corrections, if any.
Thank you very much.

PR #368, opened 2024-05-27
Update README.md

del000:
Added sound setup & tweaks page


flyrouter (changes requested):
The file specified in the link does not currently exist; corrections and additions are needed.

PR #370, opened 2024-05-27
Patch 2

del000:


flyrouter (changes requested):
Thank you for your interest in our project

It is advisable to make your corrections and additions in the Majestic section of the file
https://github.com/OpenIPC/wiki/blob/master/en/majestic-streamer.md

I think it would be more correct to write corrections to the majestic file and do it in a general form rather than specific examples like yours, since this will cause problems for beginners in case of differences.

Thank you

PR #386, opened 2024-07-15
fix dead link in README.md

RoboSchmied:
there is no file `developers.md` so a link to `contribute.md` was added instead


flyrouter (commented):
Thank you

PR #407, opened 2024-09-08
Update faq.md

iGraveD:
Configuring WireGuard to work with OpenIPC


flyrouter:
Thanks for the suggestion.
However, WireGuard is a self-sufficient application and it makes sense to put it in a separate file with the addition of a section to the menu, as is done for example by vtun.

flyrouter (changes requested):
Will be added later as a separate file

flyrouter (approved):

PR #411, opened 2024-09-22
add connect to internet operations and lower power input Update fpv-openipc-aio-mario.md

KennyPlus:
add connect to internet operations and lower power input


flyrouter:
Super, thank you

PR #414, opened 2024-09-28
Create instructions for firmware updates via internet forwarding

wildcapuchino:
This Merge Request introduces a method to update the firmware of **AIO Mario** devices over **USB** by forwarding internet traffic through the host PC. It leverages Linux-specific network commands to set up the local PC as a network bridge between the AIO board's Ethernet interface and the PC's internet connection.

This approach allows AIO Mario internet access for firmware updates even when directly connected via USB, without needing native internet connectivity.

### DIFF
--- a/en/fpv-openipc-aio-internet-forwarding.md
+++ b/en/fpv-openipc-aio-internet-forwarding.md
@@ -1,0 +1,132 @@
+## AIO Mario firmware update over usb via internet forwarding
+This method will work on Linux due to specific network commands. The key idea is to use the local PC as a network bridge betwen AIO board ethernet interface and your local internet interface.
+
+## Steps
+1. Connect AIO Mario to USB
+2. Identify your PC address. On your local PC run:
+```bash
+ip -c a
+...
+3: enx00e099fead02: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
+    link/ether 00:e0:99:fe:ad:02 brd ff:ff:ff:ff:ff:ff
+    inet 192.168.1.11/24 brd 192.168.1.255 scope global noprefixroute enx00e099fead02
+       valid_lft forever preferred_lft forever
+    inet6 fe80::de74:78fc:9af0:f031/64 scope link noprefixroute 
+       valid_lft forever preferred_lft forever
+```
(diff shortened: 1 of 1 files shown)


en/fpv-openipc-aio-internet-forwarding.md
```diff
@@ -0,0 +1,132 @@
...
+```bash
+#!/bin/bash
+
+# Get the interface name of the wireless card by stdin
+# $1 is the interface name of the AIO card
+
+USB_AIO=$1
+echo $1
+
+sudo iptables -A FORWARD -i $USB_AIO -o wlp2s0 -j ACCEPT
```
seriyps, line 68: Maybe take the internet interface as parameter as well instead of hardcoding it?

en/fpv-openipc-aio-internet-forwarding.md
```diff
@@ -0,0 +1,132 @@
...
+sudo iptables -A FORWARD -i wlp2s0 -o $USB_AIO -m state --state RELATED,ESTABLISHED -j ACCEPT
+sudo iptables --table nat -A POSTROUTING -o wlp2s0 -j MASQUERADE
```
seriyps, line 70: I think IP forwarding should be also explicitly enabled in sysctl?

wildcapuchino: You are right, port forwarding needs to be explicitly enabled in sysctl. Thanks for your comments and your time

en/fpv-openipc-aio-internet-forwarding.md
wildcapuchino: Script updated accordingly

flyrouter (approved):
Thank you

seriyps:
I tried this instruction myself - works as a charm!

One suggestion I can add to simplify it a bit is to add just one line to `forward.sh` script:

```
sudo sysctl net.ipv4.ip_forward=1
```

this will enable the forwarding up to the next reboot, but it is the same for iptables rules. It makes the instruction easier to follow because you don't have to edit `/etc/systcl.conf` and ahve separate instructio nfor debian

seriyps:
Another idea for figuring out what interface is your "default internet interface" is to run
```bash
$ ip route get 8.8.8.8
8.8.8.8 via 192.168.0.1 dev wlp4s0 src 192.168.0.194 uid 1000 
    cache 
```
it will print the interface name

PR #416, opened 2024-10-06
Created a new wiki article to explain setup of SSH using keys and no passwords

cdg123:
Although there was a section in the FAQ on how to get SSH setup for password less access it was out of date. 

So after working through how Dropbear is used for SSH I thought it was worth creating a new wiki article on how to set up the keys and to give people new to SSH a little background on what is needed and why.

Hopefully it will be of value to other people.

I updated the FAQ to point to this but have not added a specific link from the wiki menu.


flyrouter (approved):
Thank you

flyrouter:
I hope that users will read this.
Usually they write about two problems - either too little information or too much. ;)

PR #425, opened 2024-11-05
Adds warning about newer foscam models with different designation

karlitos:
So that people avoid bricking their devices


flyrouter (approved):
Thank you

PR #431, opened 2025-01-05
Translated sandbox

JohnDGodwin:
upload of the sandbox-fpv hub with translations


flyrouter:
Thank you very much, sir.
This update has been put on hold for 1-2 weeks due to the upcoming reorganization of the documentation system so as not to break the current accesses and links.

PR #438, opened 2025-01-23
Add 873XBU rf output

KennyPlus:
measure under R&S power meter,

 @ 5.805GHz, tor +/- 1.2db, input attennuator 10db


flyrouter (approved):
Thank you

PR #446, opened 2025-04-10
add note about sdcard settings with runcam default firmware

mikecarr:
Add a note about user.ini and configurators with Runcam default firmware


flyrouter (approved):
Thank you

PR #457, opened 2025-05-30
Clarify info about connector

zai1208:
The connector mentioned was the one on the mipi cable not the one that goes on the PCB, I have added that in. Please feel free to reword this.


flyrouter (commented):
Thank you

PR #458, opened 2025-07-13
Sysupgrade article was moved to Docs.

BlackFox-org:
Sysupgrade article was moved to Docs. So I removed article and added link to docs


flyrouter (commented):
Thank you. This update is currently on hold due to the upcoming reorganization of the documentation system in order not to break current accesses and links.

PR #476, opened 2026-06-20
dev-nfs-boot: document openipc-nfs-root package + standalone manual guide

widgetii:
## Summary

Expands `en/dev-nfs-boot.md` into a complete, self-contained guide to NFS root boot, covering both the hands-on manual method and the new `openipc-nfs-root` Buildroot package.

### Manual method
Rewrote the thin u-boot snippet into a full step-by-step walkthrough: UART console wiring (with the 12 V warning), saving the current u-boot env, TFTP server setup (kernel), NFS server setup (rootfs), the `bootargs`/`nfsroot` boot sequence, first login, and making it permanent / reverting to stock. This content was translated from the external Russian write-up and folded in, so the page is now standalone English with no cross-language links.

### `openipc-nfs-root` package
Documents the opt-in package workflow: why to use it, building via `contrib/make_nfsroot` or the `openipc.fragment`, the config options, server setup, u-boot bootargs (including the `ethaddr=`/DT MAC handling), read-only vs read-write modes, the no-op `ifup`/`ifdown` networking behaviour, and troubleshooting.

## Review request

@tinylabs — since this documents your NFS-root work, could you sanity-check the package section for accuracy (the `make_nfsroot` flags, config defaults, RO/RW semantics, and the bootargs example)? Want to make sure the docs match the implementation before this goes live.

### DIFF
--- a/en/dev-nfs-boot.md
+++ b/en/dev-nfs-boot.md
@@ -1,18 +1,358 @@
 # OpenIPC Wiki
 [Table of Content](../README.md)
 
-Boot device with NFS
---------------------
+## Boot device with NFS
 
-Example for HI3516EV200 device
+Booting a camera with its root filesystem on an NFS share lets you develop and
+debug **without reflashing**. Edit files on your workstation, reboot the camera,
+and the change is live. It also scales to a fleet: many cameras can boot from one
+server, each optionally served a different kernel/rootfs.
+
+There are two ways to do this:
+
+1. **[Manual method](#manual-method)** — point u-boot at an NFS share by hand,
+   using any existing firmware. Quick, no rebuild.
+2. **[The `openipc-nfs-root` package](#the-openipc-nfs-root-package)** — build a
+   purpose-made NFS-root image with stable MAC handling, a writable overlay, and
+   automatic boot-time setup (password, DNS/NTP, hostname, SSH key).
+
+---
+
+## Manual method
+
+This walkthrough boots a camera over NFS/TFTP using any existing firmware, by
+typing commands at the u-boot prompt over a UART console. Nothing in flash is
+modified unless you explicitly `saveenv`, so it is non-destructive and easy to
+revert. The worked example uses a XiongMai XM530 camera; only the SoC-specific
+values (`mem=`, load address, `mtdparts`, console name) differ between boards.
+
+You will need two services reachable from the camera — a **TFTP server** for the
+kernel and an **NFS server** for the rootfs. They can be the same machine.
+
+### 1. Connect a UART console
+
+Wire a 3.3 V USB-UART adapter to the camera's GND / RX / TX pads and open a serial
+console at 115200 baud:
+
+```sh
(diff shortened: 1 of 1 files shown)


tinylabs:
Looks good! Couple notes:

> Omit it to let a device-tree MAC win; omit both for a random MAC

Even with ethaddr= in bootargs the DT mac will win currently.

> bootargs section
> hostname= — optional; otherwise a DHCP- or build-supplied hostname is used.

This is the preferred method for setting hostname since it works across multiple cameras all mounting the same rootfs as RO. The only way it gets overridden (when present) is when we pass it via the DHCP request and the server responds with a different hostname due to a static lease. In my testing it's always been honored when using dynamic leases (on opnsense at least). For the DHCP use case I would recommend setting hostname= (in bootargs) and bootp_vci (in uboot env) to the same unique name. That makes tagging and option serving easier when configuring DHCP.

Here's a useful configuration which might be nice to add.
- On the NFS server have a single readonly rootfs per platform. 
- For each camera you can have a small overlay directory mounted with overlayfs (on the NFS server) using the common readonly root as lowerdir. 
- Mount each camera RW using the specific camera rootfs mount.
- All small tweaks go into the camera specific rootfs dir (ssh keys, majestic tweaks, etc).
- The common base stays the same and can be copied directly over when deploying a new make_nfsroot build.
- Makes it easy to separate camera specific changes from the base rootfs image.

The downside is it relies on overlayfs running on the nfs host server which is linux specific. Unfortunately we can't do this on target since the kernel NFS driver doesn't support an NFS mount on the upperdir.

Everything else looks good!

tinylabs:
One other addition: It may be good to add a small mtdparts snippet to the bootargs example. Just enough so fw_setenv can find the u-boot env partition. Without it the ssh key stuffing doesn't work and /etc/fw_env.config doesn't get created on boot which throws log warnings which confused me initially. 

I'm using this:
```
mtdparts=sfc:256k(boot),64k(env),-(unused)
```

widgetii:
Thanks @tinylabs — all four folded into 82e1155:

1. **MAC precedence** — fixed to state a device-tree MAC always wins; `ethaddr=` is only used when the DT has no MAC.
2. **`hostname=`** — now documented as the preferred method across a shared RO rootfs, with the `hostname=` + `bootp_vci` pairing tip for DHCP fleets.
3. **Per-camera server overlay** — added a "Per-camera config on the server (fleets)" subsection (shared RO base as `lowerdir` + per-device `upperdir` via overlayfs on the NFS host), including the caveat that it can't run on-target since the kernel NFS client can't use an NFS mount as an overlay `upperdir`.
4. **`mtdparts=`** — added `mtdparts=sfc:256k(boot),64k(env),-(unused)` to the bootargs example with a note on why it's needed for `fw_setenv` / SSH-key persistence, plus an updated troubleshooting row.

Also added you as a formal reviewer now that the invite's accepted. Let me know if the overlay description matches what you're running.

tinylabs (approved):
Looks good to me! I think you captured everything.

PR #478, opened 2026-08-04
memory-tuning: document the streamer's own memory settings

widgetii:
The article sizes the memory *region* thoroughly, but never says what the video pipeline puts **into** it — so the settings that actually move that number were undocumented. Adds a section covering the three of them.

**`isp.blkCnt`** — how many full frames are pooled. The biggest single lever, and the easiest to overshoot.

**`isp.memMode`** — the gap most worth closing. The name suggests it shrinks the frame pool; it does not. It changes encoder mini-buffer mode, the compressed-output buffer size, and reference-frame sharing. Worth knowing before someone tunes it expecting frames to get smaller, and worth knowing that `normal` is the thing to try when frames are dropped under a bitrate spike.

**`isp.yuvCompression`** — frame-buffer compression from a recent update. Measured 5.8 MB freed on a hi3516ev300 at 2592x1520 and 2.1 MB on a gk7205v200 at 1080p.

Two things the section is careful about, because both are easy to get wrong:

- **It does not help on every chip.** The saving comes from *reserving* less, and only some SoCs' buffer arithmetic will reserve less for a compressed frame; the rest reserve full size regardless. The table records which is which — measured on hardware, not inferred from the family name.
- **It is ignored when it would break something**, and the incompatibilities are silent ones: rotate and mirror stop the pipeline, and privacy masks are simply not drawn — a stream that looks correct while publishing the area you meant to hide. Listed explicitly so nobody has to discover that the hard way.

Every example ends in a restart. All three settings are read when the pipeline starts, and setting one through the API persists it and returns `200` while the running pipeline keeps its buffers — an easy way to conclude a setting does nothing. Verified end to end on hardware: the documented sequence takes `/proc/media-mem` from 49184 KB to 43256 KB.

Config examples use the HTTP API, following #477.

### DIFF
--- a/en/memory-tuning.md
+++ b/en/memory-tuning.md
@@ -212,16 +212,17 @@ but you will benefit from updating `mem=` to `128M` afterwards.
 The sections above size the memory *region*. This one covers what the video
 pipeline puts **into** it, and the three settings that change how much it takes.
 
-All three settings below are read when the video pipeline starts, so a change
-only takes effect after the streamer restarts:
+All three settings below are read when the video pipeline starts, so setting
+one is only half the job — the pipeline has to be rebuilt before it takes
+effect:
 
 ```bash
-/etc/init.d/S95majestic restart
+killall -HUP majestic
 ```
 
-Setting one through the API persists it and returns `200`, but the running
-pipeline keeps its existing buffers until that restart — so always check the
-result rather than assuming the call was enough.
+The API call itself persists the value and returns `200`, but the running
+pipeline keeps the buffers it already has. Always confirm the result rather
+than assuming the call was enough.
 
 All of the region's usage is visible at runtime:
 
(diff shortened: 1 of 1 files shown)


widgetii:
Thanks — the contradiction with `en/majestic-streamer.md` was real, and it pointed at a bug in the streamer rather than in the docs.

`majestic-streamer.md` was right: the HTTP API is supposed to apply a change to the running streamer. It just wasn't happening for these particular settings, so my first draft documented the symptom (reload by hand) instead of the cause. That has now been fixed on the streamer side — these settings apply through the API like any other.

So this page no longer asks for a reload, and cross-references `majestic-streamer.md` rather than restating its behaviour. The one thing kept is what a reader still needs to expect: these decide the pipeline's buffer layout, so changing one rebuilds it and the stream drops for a moment.

While re-testing the simplified instructions on hardware I also found an interaction worth having in the page, so that's added: `isp.blkCnt=4` on its own streams and answers snapshots fine, but `isp.blkCnt=4` together with compression does not — compression claims part of the same pool, so the two levers go further together than either does alone.

PR #489, opened 2026-08-23
howto-doorbell: ring several people, and drive calls over HTTP

openipc-ai:
Follows up two requests on the majestic tracker — OpenIPC/majestic#301 (ring
more than one number) and OpenIPC/majestic#302 (start and end calls from an
API) — both of which are now implemented.

## What changed

**"Ringing more than one person"** is the new section that matters. `callTarget`
takes a list, and `ringMode` picks between ringing people one at a time
(`sequential`) or all at once and keeping whoever answers first (`parallel`),
with `ringSeconds` and `ringCycles` controlling how long and how many times.

It also says plainly when *not* to use this: if you have a PBX, its dialplan
should do the forking, because it knows which handsets are actually registered
and the camera does not. The camera-side version is for direct-dial setups that
have no PBX to ask. The existing "Group ring" bullet now points both ways
instead of only at Asterisk.

**"Controlling calls over HTTP"** documents `/api/v1/sip/call`, `/hangup` and
`/status`, with a sample `status` payload, and tells people to set a password —
a camera that will phone any address it is handed is worth rather more to a
stranger than one that only shows pictures.

## One correction

This article has always told people to run `killall -HUP majestic` after editing
`sip.*`. Until the 2026-08 builds that quietly switched SIP off until majestic
was fully restarted, so the instruction did not do what it said. It works now,
and the text notes the older behaviour so anyone on an existing build knows why
their reload appeared to do nothing.

Config keys and behaviour verified on an SSC30KQ running the current build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Sv4ojZqisqvJuvtmHqXWbP

### DIFF
--- a/en/howto-doorbell-from-camera.md
+++ b/en/howto-doorbell-from-camera.md
@@ -235,7 +235,8 @@ working doorbell.
 ## Ringing more than one person
 
 A doorbell that rings one handset is a doorbell that gets missed. `callTarget`
-accepts a list — separate the entries with commas (spaces and `|` work too):
+accepts a list. Separate the entries with commas; whitespace around them is
+ignored, so write it however reads best:
 
 ```yaml
 sip:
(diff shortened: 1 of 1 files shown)


openipc-ai:
Both addressed in `c68411a`.

**1. SIP API curl auth missing** — correct, and the contradiction was real: the examples answered 401 as written. They pass `-u root:PASSWORD` now, and the on-camera case is shown separately, because a request originating on the device counts as local and needs no credentials — that is the form a button script or an on-device plugin actually uses. Verified against a camera: 401 with `WWW-Authenticate: Basic` from another host, 200 from `127.0.0.1`.

**2. Ambiguous callTarget separators** — fair. The parser does accept whitespace, but documenting it as a delimiter invites configs nobody can read. The prose now recommends commas, matching the example, and says whitespace around them is ignored, which is the part a reader needs and is equally true.

PR #490, opened 2026-08-24
docs: explain why load average reads 12-13 on SigmaStar cameras

openipc-ai:
## Why

Users adding SigmaStar cameras to Zabbix/Prometheus get overload alerts immediately and permanently. The Device Status page shows something like `load 13.35` right under `CPU 33 %`, which reads as a camera in serious trouble.

It isn't. On these SoCs a load average of ~12 (13 with audio enabled) is the **idle** value and it will never go lower.

## The mechanism

Linux counts tasks in **uninterruptible sleep (`D`)** towards the load average, not just runnable ones. The SigmaStar MI kernel modules give each pipeline port a kernel thread, and those threads wait for their next frame in uninterruptible sleep. They burn no CPU and are not stuck — they just idle in the one sleep state the load average counts.

Measured on an SSC30KQ (Infinity6E, imx335, single core, kernel 4.9.84):

```
$ cat /proc/loadavg
12.49 12.35 12.29 1/78 2452      <- 1 runnable task out of 78

$ top -b -n1 | head -3
CPU: 18.1% usr  9.0% sys  0.0% nic 72.7% idle
Load average: 12.38 12.33 12.29 1/79
```

Twelve threads in `D`, load 12.4, 72% idle. Stopping the streamer drops the count to 2 and the load decays 12.39 → 2.47 over 185 s, exactly along the 60 s EMA curve; restarting brings both back. Enabling audio adds one thread (`ai0_P0_MAIN`) and the load settles at 13.36 — which reproduces a field report's screenshot value of 13.35.

The count is set by pipeline topology, not by stream config: enabling a second video channel adds no thread.

## What the article covers

- The symptom, and the short answer that the camera is fine
- What load average actually measures, and why `D` matters here
- The thread table, and the `msleep()` note (uninterruptible by definition)
- Commands to confirm it on any camera — including the `1/78` field, which is the quickest tell
- **What to monitor instead**: CPU utilisation from `node_cpu_seconds_total`, with a PromQL expression and a Zabbix preprocessing note, plus a table of the other metrics that mean what they appear to mean
- How to spot a *real* overload above the floor
- Why it can't be fixed our side — `TASK_IDLE` exists for exactly this since Linux 4.2, but the MI modules are proprietary and out-of-tree

Filed under Troubleshooting in the README.

## Note

I linked the GitHub blob URL rather than `wiki.openipc.org` — that domain no longer belongs to OpenIPC (it resolves, but serves an unrelated commercial site, and every article path 404s). Worth a separate cleanup: `majestic-webui`'s `mj-settings.cgi` still points users there in the "Majestic is not running" alert.

A companion `majestic-webui` change adds a tooltip and a link to this article from the load line on the Device Status card.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JQXuiLPei8LC4r9S3j9KU2

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ OpenIPC Wiki
 
 - [Network does not work on hi35xx](en/trouble-network-hi35xx.md)
 - [Majestic does not work, camera reboots](en/trouble-majestic.md)
+- [Load average looks impossibly high](en/trouble-load-average.md)
 - [Sigmastar unbrick instructions](en/sigmastar-unbrick.md)
 - [Ingenic T31 unbrick with SD card](en/ingenic-t31-unbrick-with-sd-card.md)
 
--- a/en/trouble-load-average.md
+++ b/en/trouble-load-average.md
@@ -20,13 +20,19 @@ The camera itself streams normally, stays cool and does not reboot.
 
 ### Short answer
 
-**Nothing is wrong with the camera.** On SigmaStar SoCs a load average of
-roughly 12 (or 13 with audio enabled) is the *idle* value, and it will never go
-lower. Load average on Linux is not a CPU-usage figure, and on these cameras it
-cannot be used as an overload signal at all.
+**Nothing is wrong with the camera.** Load average on Linux is not a CPU-usage
+figure, and on these cameras most of what it reports comes from driver threads
+that are fast asleep.
 
-Remove any trigger you have on `node_load1` / `node_load5` / `node_load15` and
-alert on CPU utilisation instead — see [What to monitor
+Every SigmaStar camera has a *floor* — a load average it shows when completely
+idle and never drops below. On the SSC30KQ measured for this article the floor
+is about 12, or 13 with audio enabled. **The exact value is a property of the
+SoC, the firmware and the enabled pipeline, not a constant**, so establish the
+floor for your own camera rather than assuming these numbers;
+[Checking it yourself](#checking-it-yourself) shows how.
+
+Remove any trigger you have on `node_load1` / `node_load5` / `node_load15` at
+these levels and alert on CPU utilisation instead — see [What to monitor
 instead](#what-to-monitor-instead) below.
 
 ### What load average actually measures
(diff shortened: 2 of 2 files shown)


openipc-ai:
Addressed all three review findings in 5422cf9.

**1. Idle floor overgeneralised** — fair, and it contradicted the article's own mechanism. The floor's *existence* is universal on SigmaStar; its *value* follows the pipeline ports the SDK creates, so another SoC, firmware build or module set has a different one. The short answer now presents 12/13 as the measured SSC30KQ example, says explicitly that it is not a constant, and sends the reader to establish their own baseline. The mechanism section says the same where the thread table is.

**2. Blocked contention dismissed** — this one was a real defect in the advice, not just imprecision. The article closed with "if CPU idle is healthy and the run queue is 1, the load average is measuring sleeping driver threads, no matter how large it looks". A task newly wedged in `D` — failing SD card, stalled USB/network filesystem, deadlocked driver — raises the load average with **no** CPU use and an **unchanged** runnable count, so that test tells users to dismiss exactly the case where a high load average is diagnostic.

The section is rewritten around the two distinct causes, and the test is now a diff of the `D` list against a recorded baseline, which is the only thing that separates them:

```
for t in /proc/[0-9]*/task/[0-9]*; do
    [ "$(awk '{print $3}' "$t/stat" 2>/dev/null)" = "D" ] && cat "$t/comm"
done | sort | diff /tmp/loadavg-baseline -
```

`node_procs_blocked_total` is documented as a second opinion: it counts only tasks blocked in kernel I/O waits, and the vendor threads never enter it — measured 0 on a healthy camera while the load average sat at 12 — so a sustained non-zero value is a genuine I/O stall.

Both commands were verified against busybox on an SSC30KQ: the baseline captures 12 lines, an unchanged diff is silent, and a synthetic extra task shows up as `+zz_wedged_task`.

**3. Zabbix ratio underspecified** — correct, "the same two counters" was left over from an earlier draft and could not reproduce the PromQL. The denominator must sum every mode; the PromQL now does that with `sum by (instance)` on both sides instead of an `ignoring()`/`without()` pairing, and the Zabbix recipe names the master item, the two dependent items with their Prometheus patterns and preprocessing steps, and the calculated item — with a note for versions predating the `sum` function.

Also qualified the closing paragraph, which called the number "cosmetic" without limiting that to the vendor threads' contribution.

PR #496, opened 2026-08-27
majestic: what the Lite, Ultimate and FPV builds each contain

widgetii:
People keep asking how the three Majestic builds differ, so the answer now sits on the page they already read.

### What's new

A **Lite / Ultimate / FPV** section in `majestic-streamer.md`: what every build shares, a table of what the flavours actually change, how to ask a camera which one it is running (`majestic -v`), and which SoCs each flavour is published for.

The sizes in it are measured, not estimated — GK7205V200 built three times off the same commit with the same toolchain, stripped:

| flavour | binary |
|---|---|
| FPV | 496 KB |
| Lite | 917 KB |
| Ultimate | 1.24 MB |

### What was wrong

Checking every claim on these pages against the streamer turned up a fair amount of drift:

- `motionDetect.skipIn`, `nightMode.dncDelay` and `records.splitRecord` no longer exist. The real keys are `records.split` and `nightMode.monitorDelay`.
- `records.path` is a **directory** and `records.filename` the base name; the extension is appended. The documented `path: /mnt/.../%H.mp4` creates a directory of that name.
- The signal table was wrong. `SIGUSR2` places and ends a SIP call — it is not "SDK shutdown". `SIGQUIT` is the one that releases the SDK while the process keeps serving, which is what `sysupgrade` uses. The Ingenic caveat is a one-second deferral while a CGI runs, not a T21 special case.
- The snapshot parameter is `gray`, not `color2gray` — and a parameterised snapshot additionally needs HiSilicon and `jpeg.tuned`, so the documented URL fails on most cameras.
- `jpeg.size` exists and applies to both `/mjpeg` and `/image.jpg`, where the page said no such setting was provided.
- A non-root system account already reaches the media endpoints over HTTP; the page still described that as a future intention. `onvif.enabled` defaults to `true`, and the `onvif.username`/`password` pair that unlocks digest auth for older clients (ODM 2.2.x, tinyCam) was undocumented.

`majestic-config.md` also gains the sections that were missing outright — `sip`, `mdns`, `cloud`, the RTSP aliases and back-channel, the OSD anchor and offsets — and marks per-vendor keys as such, since a key the build lacks answers `404` rather than being accepted and ignored.

Plus one unclosed markdown link in `majestic-research.md`.

`majestic-encoder-tuning.md` and `majestic-plugins.md` were checked against the source too and needed nothing.

### DIFF
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -47,9 +47,9 @@ isp:
   #lowDelay: false
   #awbMode: auto                # auto|manual|day|cloudy|incandescent|
                                 # flourescent|twilight|shade|warm|custom
-  #memMode: reduction           # normal | reduction        (HiSilicon)
-  #slowShutter: disabled        # disabled|low|medium|high  (HiSilicon)
-  #dis: false                   # digital stabilisation     (HiSilicon, Ingenic)
+  #memMode: reduction           # normal | reduction        (HiSilicon/Goke)
+  #slowShutter: disabled        # disabled|low|medium|high  (HiSilicon/Goke)
+  #dis: false                   # digital stabilisation     (HiSilicon/Goke, Ingenic)
   #wdr: 0                       #                           (SigmaStar)
   #edgeGain: false              #                           (SigmaStar)
   #iqServer: false              #                           (SigmaStar)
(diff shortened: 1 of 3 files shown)


widgetii:
Both findings were valid; fixed in 09cddc4.

**1. `jpeg.enabled` (high).** Correct, and a good catch. I took "still snapshots are unaffected" from the streamer's own field hint for that key — but `memory-tuning.md` already documented the opposite, and the SoC layer settles it: with JPEG off no block is reserved and `/image.jpg` is refused **503** by name. Since ONVIF snapshot URIs and the WebUI preview both fetch `/image.jpg`, the old wording invited people to break both. Now says so on both pages and links to `memory-tuning.md` for the measured saving.

Worth noting the hint string in the streamer is itself wrong on this and contradicts the SoC layer — the WebUI shows it, so it wants fixing on the code side separately.

**2. Goke (medium).** A genuine inconsistency between my two pages. `HiSilicon` is the right side: the Goke parts build against the HiSilicon SDK, so parameterised snapshots do work there. Spelled `HiSilicon/Goke` throughout the config page, matching how the rest of the wiki writes it.

PR #501, opened 2026-09-01
docs: a factory reset takes the password with it, and scp notices first

openipc-ai:
Reported as [OpenIPC/firmware#2345](https://github.com/OpenIPC/firmware/issues/2345): after **System → Backup & Restore → Reset firmware**, SSH stopped accepting `12345`, and the reporter got back in only through the web interface.

#500 took `12345` out of the wiki, which is most of the answer. This adds the two pieces it does not cover — both of them the pieces that produced the report.

## The reset button is `firstboot`

The Danger zone button in the WebUI, `firstboot`, and `sysupgrade -n` are one operation: it erases the overlay, and the root password lives there. So the camera comes back **unclaimed**, exactly as if it had just been flashed.

`firstboot` is already documented — in `en/source-code.md` and `ru/configuration.md`, as a step during *installation*. Nobody arriving from a button labelled "Reset firmware" reads either of those.

## The doors that cannot ask for a password

The claim prompt is root's login shell, so it sits on the path of every door — including the ones with nothing to type into. `scp`, `sftp` and `ssh camera command` are refused outright until the camera has an owner:

```
This camera has not been set up yet, so it has no shell to offer.
Log in interactively (ssh root@camera, or the serial console),
or open http://camera/ in a browser, to set a root password.
```

That is what WinSCP, `scp`, `rsync`, `ansible` and any provisioning script will hit — and several guides here tell people to copy files to a camera they have just flashed. Nothing in the wiki mentioned it.

## Verification

hi3516ev300 on `nightly-20260831-bebd80a`, put into the exact post-reset state by restoring `/etc/passwd` and `/etc/shadow` from `/rom` (the pristine squashfs), so the lab camera's own configuration was not wiped:

| Door | Result |
|---|---|
| `ssh root@camera` | accepted with no password, runs the claim prompt, `passwd`, then a real shell with `/etc/passwd` repaired to `/bin/sh` |
| serial console (`busybox login`) | identical — `root`, no password asked, same prompt |
| `ssh camera id` | refused with the message above |
| `scp` | same |

So the reporter's "SSH only works after the web interface" is not what the firmware does. What was missing was anything telling them to send **no** password rather than `12345`.

## What is here

- `en/first-boot.md`, `ru/first-boot.md` — deliberately a complement to what #500 wrote rather than a second copy: the reset, the refused doors, provisioning, and `system.unsafe: true` for cameras deliberately run with an empty password for open RTSP.
- Links from both web-interface pages, both `firstboot` mentions, the FAQ's SSH and `scp` entries, and the table of contents.

No firmware change is proposed: nothing is broken.


openipc-ai:
Thanks — I chased this one down, and the premise turns out to be inverted, though the concern behind it was real and came from this wiki.

**`-n` does not upgrade anything.** In `general/overlay/usr/sbin/sysupgrade`, `-n | --wipe_overlay` sets `clear_overlay=1` and nothing else. `remote_update` is set only by `-k`, `-r`, `--url=`, `--channel=`, `--build=` and `--archive=`, and the download is guarded on it:

```sh
[ "1" = "$remote_update" ] && download_firmware
```

A bare `sysupgrade` sets none of `clear_overlay`, `update_kernel` or `update_rootfs`, so it never reaches that line at all:

```sh
[ "1" != "$clear_overlay" ] &&
[ "1" != "$update_kernel" ] &&
[ "1" != "$update_rootfs" ] &&
        echo_c 37 "\nTry '$(basename "$0") --help' for options." &&
        exit 0
```

Measured on a hi3516ev300 running `nightly-20260831-bebd80a`:

```
# sysupgrade
OpenIPC System Updater v1.0.62

Vendor  hisilicon
SoC     hi3516ev300
Model   hi3516ev300
Kernel  17:57:17 2026-08-31
RootFS  master+bebd80a, 2026-08-31

Try 'sysupgrade --help' for options.
# echo $?
0
```

**But the evidence cited is a real defect, in `en/sysupgrade.md`.** That page opens with "By default, running sysupgrade will attempt to download the latest software for your camera model", then contradicts itself four lines later with "For newer firmware, run `sysupgrade -k -r`". Anyone reading it top to bottom — a model or a person — comes away believing a reset flashes firmware. That is exactly the reading this finding arrived at, so the finding is doing its job even though its conclusion about `-n` is wrong.

Fixed in cab5d2d:

- `en/first-boot.md` and `ru/first-boot.md` now say outright that `-n` on its own writes no firmware, name the six options that do, and note that a bare `sysupgrade` only prints the camera's identity.
- `en/sysupgrade.md` leads with `sysupgrade -k -r` and describes what the bare command actually does, instead of the stale default.

PR #503, opened 2026-09-04
Majestic: automatic day/night needs no thresholds now, and the lamp can dim

openipc-ai:
Documentation for the night-mode features that land with the next majestic nightly:

- **Auto day/night detection** rewritten: `lightMonitor: true` alone now selects an exposure-based mode with nothing to calibrate; the three-way source chain (sensor pin → raw thresholds → automatic); the tuning knobs in portable gain multiples; the `night_mode_source` gauge and the live decision chart on **Settings → Day / Night**; and the two park switches (`irCutEnabled`/`backlightEnabled`) that keep wiring configured while an actuator stays still. The raw-threshold instructions remain as the documented legacy mode, with the per-SoC warning sharpened — the gain *scales* differ between vendors, not just the values.
- **New section: PWM backlight** on the EV200/EV300/GK7205V200/V500 family — the in-camera version of the sandbox `devmem` backlight scripts: channel/frequency/duty keys, the ambient-tracking behaviour, why the lamp cannot talk the camera back into day, and the `night_light_duty` gauge.
- The old [auto-night-mode-without-light-sensor](en/auto-night-mode-without-light-sensor.md) article's banner now names both generations of its replacement and keeps the scripts for older firmware.

### DIFF
--- a/en/auto-night-mode-without-light-sensor.md
+++ b/en/auto-night-mode-without-light-sensor.md
@@ -1,11 +1,17 @@
 # OpenIPC Wiki
 [Table of Content](../README.md)
 
-**Attention, this instruction is outdated since summer 2024.** Majestic switches
-day/night by itself from the sensor's analog gain — see
-[auto day/night detection](majestic-streamer.md#auto-daynight-detection) for the
-two settings that replace these scripts, and set them from **Settings → Day /
-Night** in the web interface.
+**Attention, this instruction is outdated twice over.** Since summer 2024
+majestic can switch day/night by itself from the sensor's analog gain (the two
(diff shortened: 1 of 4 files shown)


openipc-ai:
Both findings addressed in d963e90 — each was the wiki contradicting itself: (1) ircut-filter.md's banner list no longer calls 'a light monitor with nothing to watch' a broken configuration — on current firmware that state selects the automatic mode, and the text now names the two cases that still earn the warning (older firmware, or a SoC reporting no exposure state); (2) majestic-config.md's nightMode reference block now carries the park switches, the four automatic-mode knobs, and the PWM backlight keys with their platform limitation, in the file's own comment style.

PR #504, opened 2026-09-04
Majestic: motion can record now, and the recording keys have caught up

openipc-ai:
The **Motion detection** section said a script was the only thing a motion event could drive — because until September 2026 it was. Somebody read exactly that and asked on the tracker how to record on motion *"without /usr/sbin/motion.sh????"* ([majestic#261](https://github.com/OpenIPC/majestic/issues/261)); somebody else asked the same thing about a birdbox camera fifteen months earlier ([#253](https://github.com/OpenIPC/majestic/issues/253)); the original request has been open since 2023 ([#136](https://github.com/OpenIPC/majestic/issues/136)). All three are closed now, and the article was the last thing still telling people it could not be done.

## Recording on motion

A new section under Motion detection, with the config and — more importantly — the one setting that decides whether it does what people expect.

A clip can only begin at a keyframe. The camera holds the last `preRollSec` of video in RAM so the clip can start *before* the detector fires, which is the whole point of the feature for a nest box or a doorway. But if `gopSize` is longer than that window there is usually no keyframe among the seconds being held, and the run-up is discarded.

With the shipped defaults — `gopSize: 30` against `preRollSec: 5` — a keyframe falls inside the window about **one event in six**. That is worse than never, because the first clip somebody checks may be the one that worked. So the section says to set `gopSize` at or below `preRollSec`, and shows the measurement: with both at 5, the clip's first four seconds are an empty room and the person walks in at second five.

## What survives a power cut

None of this was written down anywhere. The card is committed on a timer (`records.syncSeconds`) rather than per frame; a torn or reused-cluster tail can be trimmed with `majestic --repair`; and `/metrics/records` is majestic's own verdict on the card, which the filesystem cannot give you — a card can be mounted read-write with room on it and still be taking nothing.

## Config example

`records` had fallen a long way behind: it gained `mode`, `preRollSec`, `postRollSec`, `minClipSec`, `purgeSeconds`, `syncSeconds`, `fragmentMs` and `fragmentBytes`; `hls` gained `segments`.

Two things noted while adding them, because both surprise people:

- **`split` now cuts at the next keyframe after it falls due**, so a clip can run up to one `gopSize` longer than the number set. It has to — a clip cut anywhere else opens on frames whose references are in the previous file.
- **There is no `motionDetect.exclude` key.** `roi` says where motion counts, and nothing says where it does not. Worth stating outright: the parser accepts such a line and reads nothing from it, so it comes back in `/api/v1/config.json` looking as though it took.

cspell clean.

### DIFF
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -229,13 +229,16 @@ nightMode:                      # see en/ircut-filter.md for how the filter is
   #backlightPwmMin: 10          # duty floor, % — LEDs have an ignition threshold
   #backlightPwmMax: 100         # duty ceiling, %
 
+# Motion runs /usr/sbin/motion.sh, and — with records.mode below — can record a
+# clip per event. There is no "exclude" key: roi says where motion counts, and
+# nothing says where it does not.
 motionDetect:
   enabled: false
   visualize: false              # draw a box around what moved. On SigmaStar this
                                 # and osd.privacyMasks cannot both be drawn; if
                                 # masks are set the masks are kept and the boxes
                                 # are not drawn
-  debug: false
+  debug: false                  # a log line per detection — several a second
   #roi: 1854x1304x216x606,1586x1540x482x622
   #sensitivity: 3               # 0-8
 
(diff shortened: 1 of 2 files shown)


openipc-ai:
Fixed in 88cca65 — the review is right, and the mistake was mine twice over.

`gopSize: 30` is not a shipped default. Majestic's built-in is `1.0` (`video_gop_size()` falls back to it when the key is absent) and the firmware images write `gopSize: 1` explicitly. 30 was the lab camera's own setting, which I read as if it were stock. Since 1 is already well below any `preRollSec` worth setting, a camera nobody has reconfigured records the whole run-up with no configuration at all — the opposite of what the section said.

The measurement itself stands, it just belongs to a different camera: at `gopSize: 30` against `preRollSec: 5`, three consecutive events kept three seconds, one second and nothing. That is now framed as what happens when `gopSize` has been *raised* for bandwidth, which is the only route to the problem, and the minimal example no longer sets `video0.gopSize` since the default needs no help.

The warning majestic logs was never wrong — it only fires when the run-up buffer was full and no fragment in it opened at a keyframe, so it cannot reach a camera at the default.

Same correction is going into `docs/recording-on-motion.md` and a source comment in majestic, and onto the two issue threads where I repeated it.

PR #508, opened 2026-09-05
Majestic: HLS runs beside recording now, and describes it

openipc-ai:
Three things changed for anyone using HLS, and the configuration reference still said the opposite of the first.

**HLS and records are no longer mutually exclusive.** The comment above the `hls:` block promised that turning records on would turn HLS off. That was true and is not any more, and running both is now the better configuration rather than merely a permitted one.

**`hls.enabled` applies without a restart**, in either direction, and recording carries on across the change.

**With recording on, the playlist describes byte ranges of the clip being written** rather than holding a second copy of every segment in RAM. Two consequences an operator can see:

- `malloc_hls_alloc_bytes` in `/metrics` reads **0** instead of the tens of megabytes a long GOP at a high bitrate costs;
- the live edge is **about a second** behind instead of a whole GOP, because the playlist carries `EXT-X-PART` entries for the segment still being written.

With records off, or `records.key` set — which scrambles the clips, so their bytes are not what a player needs — it falls back to keeping segments in memory. That is the old behaviour, and the expensive way round.

The new **Live HLS** section in the streamer guide also covers blocking playlist reload (`_HLS_msn`/`_HLS_part`, the 400 and 504 cases), and says plainly that **segment length still follows `video0.gopSize`** even though the live edge no longer does. A camera serving players without low-latency support wants a shorter GOP than one that is only recording, and that is not obvious from either setting on its own.

Every number quoted was measured on a lab hi3516av300 rather than derived from the code. Kept to configuration keys, endpoints, playlist tags and metric names — nothing about how any of it is implemented.

### DIFF
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -328,11 +328,19 @@ watchdog:
 # card, which costs no memory and puts the live edge about a second behind.
 hls:
   enabled: false
+  #adaptive: false              # offer both encoder channels as two variants and
+                                # let the player choose. Only the recorded channel
+                                # comes from the card, so the second one is held in
+                                # RAM: this is the setting that costs memory
   #segments: 4                  # finished segments the playlist offers, 2-8.
                                 # Only a memory cost when there is no recording
                                 # to describe — with records off, or records.key
                                 # set, each segment is held in RAM instead.
                                 # Nothing is allocated while HLS is off
+                                #
+                                # records.mode: motion turns this off: nothing
+                                # is written between detections, so there would
+                                # be no live stream to serve
 
 mdns:
   enabled: true                 # answers for openipc.local and <hostname>.local
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -323,12 +323,16 @@ watchdog:
   enabled: true
   timeout: 300
 
-# HLS and records are mutually exclusive — turning records on turns this off.
+# Applies without a restart, and runs alongside records rather than instead of
+# it — with recording on, the playlist describes byte ranges of the clip on the
+# card, which costs no memory and puts the live edge about a second behind.
 hls:
   enabled: false
-  #segments: 4                  # finished segments held for clients, 2-8. Each
-                                # is a copy of that much video in RAM, so this
-                                # is memory; nothing is allocated while HLS is off
+  #segments: 4                  # finished segments the playlist offers, 2-8.
+                                # Only a memory cost when there is no recording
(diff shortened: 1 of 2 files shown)


openipc-ai:
Both review findings addressed in 0f89e6d, and the first one turned out to be a defect in majestic rather than a gap in the page.

**Motion mode.** The reviewer is right that nothing is written between detections, and the consequence was worse than the page failing to mention it: because the recorder still reported itself healthy, the in-memory fallback stayed on standby and the live playlist simply stopped advancing until something moved. That is now refused at the configuration level — `records.mode: motion` turns `hls.enabled` off and says why, and the settings page shows the switch as unavailable — so the page documents a rule an operator will actually meet rather than a fallback that did not happen.

**`records.substream`.** Also right, and it was a straight contradiction: HLS describes whichever channel is being recorded, so the `gopSize` that sets segment length is `video1.gopSize` when the substream is the one being written. The page said `video0` unconditionally.

Also added, since it shipped after this branch was opened: `hls.adaptive`, which offers both encoder channels as two variants. Its memory cost is stated plainly — only the recorded channel can be described out of the clip on the card, so the second variant is held in RAM and costs roughly one GOP of that channel per segment kept.

openipc-ai:
Both findings were right, and both are now resolved — the second of them turned out to be a defect in the camera rather than a gap in the page.

**Motion mode.** Correct that nothing is written between detections. The consequence was worse than the page failing to mention it: because the recorder still reported itself healthy, the in-memory fallback stayed on standby and the live playlist simply stopped advancing until something moved. That combination is now refused outright — turning motion recording on turns `hls.enabled` off and logs why — so the page documents a rule an operator will actually meet rather than a fallback that did not happen.

**`records.substream`.** Also correct, and it was a straight contradiction: HLS describes whichever channel is being recorded, so the `gopSize` that sets segment length is `video1.gopSize` when the substream is the one being written. The page said `video0` unconditionally.

Two further changes since that review:

- `hls.adaptive` is documented, with its memory cost stated plainly — only the recorded channel can be described out of the clip on the card, so a second variant is held in RAM and costs roughly one GOP of that channel per segment kept.
- The sentence about the settings page said it shows the HLS switch as "unavailable". It does not: the form draws the requirement as a note under the control and leaves it editable, because the setting is a legitimate thing to want and starts working the moment the recorder leaves motion mode. Describing it as greyed out would send someone looking for a control that is not there.

Rebased on master and mergeable.

PR #509, opened 2026-09-05
Majestic: idle suspension of the sensor and ISP, and a restart that was never needed

openipc-ai:
Two unrelated things, one commit each.

## `isp.suspendWhenIdle` had no documentation at all

It ships in tonight's nightly (OpenIPC/majestic#298) and the wiki said nothing about it, which had also quietly dated three other pages.

**New section on the streamer page**, placed with the other sensor-side behaviour next to day/night and the backlight. The numbers in it are draw measured at the PoE port on a lab hi3516ev300 + IMX335, sixty samples a state:

| state | power drawn | die temperature |
| --- | --- | --- |
| both encoders streaming + audio | 2.06 W | 62.1 °C |
| both encoders disabled, front end still running | 1.77 W | 56.2 °C |
| **idle-suspended** | **1.01 W** | 49.5 °C |

That is the point of the feature and the reason it earns a section rather than two lines in the config reference: suspending halves the draw, where switching the encoders off — everything that was possible before — accounts for less than a third of the saving. Temperature understates it badly, so the section says to judge by current.

Four behaviours are documented because each one will otherwise surprise somebody: waking costs about two seconds while auto-exposure reconverges; a snapshot poll deliberately cannot wake the camera but a connected `/mjpeg` viewer can; the ISP gauges leave `/metrics` while the sensor is asleep; and it saves power, not memory.

**Three consequences elsewhere:**

- The JPEG section gave `jpeg.enabled` as *the* reason `/image.jpg` answers `503`. There are two now, with different remedies.
- The auto-night-mode page teaches readers to look for `isp_again` on `/metrics` — one of the gauges omitted while suspended. A note beats leaving someone hunting for a missing line.
- Memory tuning could reasonably leave a reader expecting a suspend to free memory. It does not; the pool is sized when the streamer starts. That paragraph also picks up the fact that `videoN.enabled` is applied live now and no longer costs every other viewer their stream.

## The reference structure never needed a restart

Unrelated to the above and older than it, which is why it is a separate commit.

The encoder-tuning page said a config reload was not enough for `refEnhance` and `refPred`, and that Majestic had to be restarted or the camera rebooted. Neither is true: setting either key through the API restarts the video pipeline on its own and the new value is picked up as encoding starts again — which is exactly the moment the same paragraph says it has to happen. Checked on hardware; the majestic process is never restarted.

The page now names the cost that is real — the streams drop and come back, so viewers see a reconnect — and points at the bitrate check it already documents for confirming the setting took. The paragraph's first sentence was correct and is kept.

### DIFF
--- a/en/auto-night-mode-without-light-sensor.md
+++ b/en/auto-night-mode-without-light-sensor.md
@@ -66,6 +66,11 @@ After stopping the script, you can run `/usr/sbin/autonight.sh` manually in a te
 Metrics are displayed at the `/metrics` endpoint in the web interface.
 
 _The current analog gain value is displayed in `isp_again`:_
+
+> On HiSilicon and Goke with `isp.suspendWhenIdle` switched on, the ISP
+> gauges — `isp_again` among them — are omitted while the camera is
+> idle-suspended, because there is no running ISP to read them from. They
+> come back with the picture.
 ```
 # HELP isp_again Analog Gain
 # TYPE isp_again gauge
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -56,6 +56,9 @@ isp:
   #pad: 0                       #                           (SigmaStar)
   #yuvCompression: auto         # HiSilicon gen 3/4 only; saves ~1/3 of the main
                                 # frame buffer, ignored with rotate/mirror/masks
+  #suspendWhenIdle: false       # HiSilicon/Goke: stop the sensor and ISP when
+                                # nothing wants frames -- see majestic-streamer.md
+  #suspendIdleSeconds: 5        # grace period before it does, 1-300
   # Manual exposure and gain. Each is skipped when absent, so leaving it out
   # keeps the automatic behaviour.
   #exposure: 0                  #                (HiSilicon/Goke, SigmaStar, Ingenic)
--- a/en/majestic-encoder-tuning.md
+++ b/en/majestic-encoder-tuning.md
@@ -199,14 +199,18 @@ answers `404` rather than accepting a setting nothing would apply.
 
 The reference structure is programmed between encoder channel creation and the
 start of encoding; the SDK ignores a later call. That does *not* mean you have
-to restart anything. Setting either key restarts the video pipeline on its own
-and the new value is picked up as encoding starts again, so this is enough:
+to restart anything. Setting either key is enough on its own:
 
 ```
 curl 'http://localhost/api/v1/set?video0.refEnhance=1'
 ```
 
-Majestic is not restarted and the camera does not reboot. The real cost is that
-the streams drop and come back, so viewers see a reconnect — do it when that is
-acceptable, not while something is recording. Confirm the new setting took
-effect with the bitrate check above rather than assuming it did.
+Majestic is not restarted and the camera does not reboot. What you will see is
+the streams drop and come back, so viewers reconnect — do it when that is
+acceptable, not while something is recording.
+
+Read the value back afterwards rather than assuming it was taken:
+
+```
+curl 'http://localhost/api/v1/config.json' | grep -A2 refEnhance
+```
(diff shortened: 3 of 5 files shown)


openipc-ai:
Reviewed under the gates from #510, now on `master`. Five of seven were right and are fixed; two are declined with reasons.

## Fixed

**Pipeline restart mechanism disclosed.** The page explained that setting a key restarts the video pipeline and that the value is taken as encoding starts again. The reader needs neither: setting it is enough, and what they will observe is streams dropping and reconnecting. Trimmed to that.

**Frame-pool lifecycle disclosed.** Same shape. "A suspended camera holds the same memory as a running one" is the entire operator-facing fact; when the pool is sized and what a suspend does not hand back is design detail.

**Snapshot guarantee now false** — and this is the one we should have caught ourselves. The snapshot section says a plain `/image.jpg` "always works". True when written, made false by a feature documented three hundred lines below it in the same pull request. That is precisely the half-updated enumeration `.pr_agent.toml` warns about, committed alongside the warning. Now points at the suspension section and notes it is off by default.

**Firmware coverage is unstated.** Fair: an operator on older firmware would set two keys their build ignores. The section now says when they arrived and that a build without them answers `404`, so a missing key reads as old firmware rather than unsupported hardware.

**Bitrate check cannot verify change.** Also fair — that check distinguishes something else, so it could not confirm a `refEnhance` change. It now reads the value back through the API, which does.

## Declined

**"ISP internals published".** `ISP` is industry vocabulary for the image signal processor, not a name internal to any one product — and here it is also the configuration namespace an operator types: `isp.suspendWhenIdle`, `isp.antiFlicker`, `isp.exposure`, all long-standing public keys. The metric names it refers to are `isp_again` and `isp_exptime`, which is what `/metrics` actually emits. Refusing the word would make the page unable to name the settings it documents. Where the *mechanism* was described in the product's own terms, that was finding 2, and it is fixed.

**"Private tracker reference disclosed".** `OpenIPC/majestic#298` is the public tracker — that is the issue this work came from, it is open to anyone, and it is where the feature was requested and answered. Nothing about it is private.

Worth noting what this round demonstrates: the gates were merged an hour ago and the first branch they were pointed at was this one, which had already passed a clean review before they existed. It failed three of them. Two were real disclosure and one was a factual contradiction, none of which the pre-gate review had raised.

openipc-ai:
**"Power saving arithmetic is wrong" — correct, and the best catch of the round.** The table rounds to two decimals; the deltas beside it were computed from the unrounded means, so subtracting 1.01 from 1.77 gives 0.76 where the sentence said 0.77, and the two components summed to more than the stated total.

Fixed to 0.76. The measurement is unchanged and so is the conclusion — the suspend is still worth about two and a half times what stopping the encoders is — but a page whose own numbers do not reconcile invites a reader to distrust the ones they cannot check, which is most of them. I have also corrected the same figure in the public issue comment where these numbers were first published.

Everything else in the current list is carried forward from earlier runs rather than re-derived: the two disclosure findings and the three correctness ones were fixed in `2ed3daf`, and the two declines (`ISP` as vocabulary, and the tracker being public) are answered above. Qodo appears to re-list prior findings without re-checking them, so the counts do not fall as they are addressed.

openipc-ai:
**"Lab hardware identity disclosed" — declined.** The gate's success criterion names this exact form as the required one:

> A measurement or verification is described by hardware class and configuration — **the SoC, the sensor**, the stream settings, how many samples over how long — which is what makes the result reproducible by someone else.

`Hi3516EV300 + IMX335` is a SoC and a sensor: a specification anyone can buy, not the identity of a machine. The failure criterion refuses a lab host's name or domain, an address on the lab network, a device-specific login — things that reach *our* hardware. Removing the SoC and sensor would leave "we measured a camera somewhere and got 1.01 W", which no reader could check or reproduce, and would make every measurement on this wiki unciteable.

That said, the finding is useful evidence about the rule rather than the page. I wrote on #510 that "a gate that refuses the repository's own subject matter is worse than no gate, because it trains people to override it" — and here is that gate, one hour old, doing exactly that to the first measurement it met. The failure criterion opens with "something identifying the project's own test hardware", and "our test hardware" can be read as the SoC we happen to own rather than as a specific machine. It should say machine.

I am not folding that into this pull request, which is documentation. It wants a one-line clarification to the gate as a follow-up, and I would rather it be reviewed as a change to review policy than buried here.

PR #510, opened 2026-09-05
Review gates: make the disclosure rule a check instead of a promise

openipc-ai:
Qodo already reviews pull requests in this repository, but it has no compliance rules here, so it has nothing to check disclosure against.

That gap showed up in #509. A page documenting a closed component went in describing how that component is *built* — its internal subsystem boundary, the vocabulary its own source uses for itself, the mechanism underneath, and a log line quoted as something to grep for. Qodo reviewed the pull request and said nothing about any of it, correctly, because nothing had told it to look. A human caught it before merge. That is not a control, and it is the one class of mistake that cannot be undone afterwards.

## Two gates

Mirrors the pair [OpenIPC/majestic-webui](https://github.com/OpenIPC/majestic-webui) adopted for the same reason, pointing the other way: that repository is public and must not reference a private one; this repository is public and must not explain what a private one is made of.

**Documents behaviour, not the internals behind it.** Says what a page about a closed component *may* state — configuration keys with defaults and ranges, endpoints and status codes, metric names, measured figures with the method used to get them, and the behaviour a reader will meet, caveats included. That is both the safe half and the more useful half: internals are the part that changes without notice, so a sentence resting on one goes stale while the behaviour it was explaining stays put.

**No lab environment details.** Keeps hostnames, addresses and credentials for the hardware that verifies these claims out of a public tree, while still asking for the SoC, the sensor and the sample count — which is what makes a measurement reproducible by someone else.

Both apply to a pull request's title and description as well as its diff. A description is as public as the page, and neither can be taken back: a commit message on a public branch cannot be rewritten, and editing a merged description leaves the original in its edit history.

The disclosure criterion deliberately gives no examples, and says so in the text. Writing one down would be the disclosure it prohibits — a trap the sibling repository fell into in the very commit that introduced its own version of this rule, which is a fair measure of how easy the mistake is.

## `.pr_agent.toml`

Carries the half that is judgement rather than pass/fail: review priorities with disclosure first, then the failure this wiki actually accumulates — a confident instruction nobody retested after the software changed. That is how a page came to insist on a restart that had stopped being necessary, also fixed in #509.

It also lists what *not* to report, so a documentation review is not spent on spelling variants, table alignment, or the `ru/` pages lagging behind `en/`.

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -60,8 +60,9 @@ something the pull request did not set out to document.
 
 compliance_user_guidelines = """
 Apply pr_compliance_checklist.yaml literally, to the lines the diff adds or changes AND
-to the pull request's own title and description — the disclosure gate is about what
-becomes public, and a pull request description is as public as the page.
+to the commit messages on the branch, the pull request's title and its description —
+the disclosure gate is about what becomes public, and all of those are as public as
+the page. A commit message is the one that cannot be edited afterwards at all.
 
 Do not raise a finding for pre-existing text that the diff merely moves or rewraps.
 """
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -1,0 +1,67 @@
+# Qodo code-review configuration.
+#
+# The binary hard gates live in a companion file at the repo root:
+#   pr_compliance_checklist.yaml
+#
+# Reference: https://docs.qodo.ai/install-and-configure/configuration-overview/configuration-file
+
+[github_app]
+pr_commands = [
+    "/agentic_describe",
+    "/agentic_review",
+]
+
+[review_agent]
+comments_location_policy = "both"
+
+# 1 = all severities inline, 2 = high+medium, 3 = high only.
+# 2 keeps disclosure findings inline without burying a documentation diff in nits.
+inline_comments_severity_threshold = 2
+
+issues_user_guidelines = """
+This repository is the OpenIPC wiki: user-facing documentation for camera firmware,
+the software that runs on it, and the hardware it runs on. Readers are operators and
+integrators, not the people who build it. Pages are Markdown under en/ and ru/.
+
+**This repository is PUBLIC, and not everything it documents is.** Majestic in
+particular is developed in a repository that is not open. Treat a page that explains
+how a closed component is built, rather than what it does, as a finding in its own
+right — see the disclosure gate in pr_compliance_checklist.yaml, the highest-priority
+check here and the only one that cannot be fixed after merge.
+
+Prioritise, in order:
+
+1. Disclosure. Internals of a component whose source is not public, and any detail
+   identifying the lab hardware used to verify a claim. Permanent once merged: a
+   commit message on a public branch cannot be rewritten, and editing a merged pull
+   request's description leaves the original visible in its history.
+
+2. Claims a reader cannot check, and claims that are simply wrong. This wiki's
+   recurring failure is a confident instruction nobody has retested since the software
+   changed — most often a stated requirement to restart or reboot for something that
+   now applies on its own, or a setting described with a default it no longer has. A
+   page that tells the reader how to confirm a result for themselves is worth more
+   than one that asserts it.
+
+3. Completeness where behaviour has more than one cause. When a page enumerates why
+   something happens — why an endpoint returns a particular status, why a feature is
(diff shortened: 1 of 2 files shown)


openipc-ai:
The four findings above are real, but they are against `1d88ff7`, which is no longer this branch.

They are worth reading anyway, because of *why* they were there. I cut this branch from a local `master` that still carried the unsanitised first draft of #509, so the diff Qodo reviewed contained five documentation files it should never have touched — and every one of the four findings is the new disclosure gate correctly firing on that draft: an internal classification, a verbatim log line, the suspend mechanism, and buffer-lifecycle detail. The gate caught, on the pull request that introduces it, precisely the class of mistake it was written for. That is a better argument for merging it than anything in the description.

The branch is now rebased onto `master` and contains only the two configuration files. #509 carries the documentation, sanitised, and currently reports clean.

openipc-ai:
Addressed.

**"Lab gate blocks valid docs" — correct, and the only finding that applies to this branch.** The criterion refused any real address, credential or default password, which is a large part of what this wiki exists to document: `192.168.1.10` appears 68 times in `en/`, and there is a security section about the password older firmware shipped with whose entire point is telling people to change it. The gate would have blocked that page and the next factory-reset walkthrough with it.

It now distinguishes the two things it was conflating. Addresses and logins belonging to the *product* — what a camera ships on, what a fresh device accepts, what a reset restores, the example ranges pages use to show a command — are explicitly allowed. What is refused is something identifying the *project's own* machines: a lab host, an address on the lab network offered as somewhere a reader could reach, a login tied to one specific device rather than a documented default.

Worth noting the shape of that mistake, since it is the second time in this pair of pull requests: the first draft of this gate was too broad in exactly the way the sibling repository's escaping gate was too narrow — a rule confident enough to fail correct work. A gate that refuses the repository's own subject matter is worse than no gate, because it trains people to override it.

**Findings 2–5 do not apply to this branch.** They are carried over from the run against `1d88ff7` and point at `en/majestic-encoder-tuning.md`, `en/majestic-streamer.md` and `en/memory-tuning.md` — none of which are in this pull request, which contains only the two configuration files. They were valid against that stale commit, and the text they name has already been removed in #509.

openipc-ai:
**"Commit messages bypass gates" — correct, and the sharpest of the findings.** The header argued that a commit message on a public branch cannot be rewritten, and then scoped both gates to the diff, the title and the description: everything except the artefact that argument was about. Internals disclosed only in a commit message would have passed a gate whose own reasoning explains why that is the worst possible place to put them.

Both gates and the reviewer guidance now name commit messages alongside the title and the description. A description can at least be edited, however visibly; the message cannot be touched once the branch is public.

**"Lab gate blocks valid docs" was addressed in the previous push** and the text at that location no longer says what the finding quotes — product-owned addresses and logins are now explicitly allowed, and only the project's own test machines are refused.

**The remaining findings point at `en/majestic-encoder-tuning.md`, `en/majestic-streamer.md` and `en/memory-tuning.md`,** none of which are in this pull request; it contains two configuration files. They are carried over from the first run, when this branch was mistakenly cut from a local `master` holding the unsanitised draft, and the text they name is already gone in #509.

openipc-ai:
**"Non-page disclosures bypass gate" — correct, and the same shape of hole as the last one.** The header claims the gates cover everything the merge makes public; the criterion then said "a page", which is narrower than the diff it claims to govern. This wiki commits images as well as prose, and a terminal capture or an editor window can carry considerably more internals than any paragraph would.

The criterion now names the artifact class rather than the file type — a page, an image or other committed artifact, a commit message, the title, the description — and the success criterion says outright that a screenshot is held to the same standard as the text beside it.

That is the third finding in a row where the rule was narrower than its own stated intent: first the lab gate refusing the product's own defaults, then commit messages excluded by a header arguing they are the least recoverable place of all, now this. Worth recording, because the pattern is not carelessness in any one sentence — it is that a criterion gets written against the example in mind rather than the category, and the gap only shows when something outside that example comes along.

**Everything else in the list is settled:** "Commit messages bypass gates" and "Lab gate blocks valid docs" were fixed in the two previous pushes, and the four `en/*.md` findings belong to a stale run against a mis-based branch — this pull request contains two configuration files. Qodo appears to carry findings forward across runs rather than re-deriving them, so the count will keep including them.

PR #512, opened 2026-09-05
docs: what encrypted recordings protect, and how to open one afterwards

widgetii:
Majestic gained real at-rest encryption for SD recordings (OpenIPC/majestic 781c635cf): CENC fragmented MP4 with three key modes, replacing the `records.key` XOR that named files `.enc`.

This adds `en/recording-encryption.md` — which theft each mode survives, what to put in `majestic.yaml`, the `--otp-burn` provisioning for chip mode, and the exact `records_key.py` + `openssl` + `ffmpeg` recipes for getting a clip back on your own machine. Linked from the table of contents.

It also fixes the `records:` block in `en/majestic-config.md`, which still documented `key: ""` as the XOR obfuscation and named the output `.enc`. That option no longer exists; the block now lists `encryption`, `key`, `publicKey` and `chipSlot`.

The three things a reader most needs and would otherwise learn the hard way: `passphrase` does not protect a stolen camera, `chip` does not either, and a `chip` clip with no `publicKey` beside it dies with the board.

cspell clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ OpenIPC Wiki
 - [Majestic encoder tuning](en/majestic-encoder-tuning.md)
 - [Majestic usage research](en/majestic-research.md)
 - [Majestic plugins](en/majestic-plugins.md)
+- [Encrypted recordings on the SD card](en/recording-encryption.md)
 - [Web interface](en/web-interface.md)
 - [Factory reset and the unclaimed camera](en/first-boot.md)
 - [Upgrade firmware](en/sysupgrade.md)
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -237,7 +237,14 @@ records:
   #substream: false             # record video1 instead of video0
   #notime: false                # ignore filename, number files 00000.mp4 upward
   #audioCodec: ""               # mp3 | aac | opus; empty follows audio.codec
-  #key: ""                      # XOR-obfuscate the payload; names files .enc
+  #encryption: none             # none | passphrase | chip | pubkey -- see
+                                # en/recording-encryption.md; a mode the camera
+                                # cannot honour stops recording, it never falls
+                                # back to writing in the clear
+  #key: ""                      # the passphrase, for encryption: passphrase
+  #publicKey: ""                # PEM public key: the only slot in pubkey mode,
+                                # a recovery key in the other two
+  #chipSlot: 1                  # which OTP key slot chip mode uses (1-3)
 
 outgoing:
   enabled: false
--- a/en/majestic-config.md
+++ b/en/majestic-config.md
@@ -242,9 +242,9 @@ records:
                                 # cannot honour stops recording, it never falls
                                 # back to writing in the clear
   #key: ""                      # the passphrase, for encryption: passphrase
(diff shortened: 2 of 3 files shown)


widgetii:
Thanks — five of the seven were real, and are fixed in a2e2bc2. Two I've handled differently, with reasons.

**Fixed as reported**

- **(4) Lost passphrase marked readable.** Correct, and it was the worst thing on the page: the row conflated the camera dying (where the passphrase you still hold opens everything) with the passphrase going (where nothing does). It is two rows now, and losing the passphrase or the private key says **lost** in every column that deserves it.
- **(5) Running pubkey camera.** Also correct. `pubkey` protects the clips already closed on the card; the clip being written has its key in RAM in every mode. The bullet now says so, and the sentence under the table is qualified the same way.
- **(6) and (7) Recipes omit `records.enabled`.** Added. Following the chip recipe as written would have burned an OTP slot — irreversibly — and then recorded nothing, which is the worst version of this mistake.
- **(1) Source paths.** Removed, and worth more than the rule it cites: Majestic's source tree is not public, so both links were 404s as well as a leak of its layout. The tool is named without a path now.

**Handled differently**

- **(2) Cryptographic internals.** Trimmed the parts an operator cannot act on — the KDF's parameters and round count are gone. The rest stays deliberately. That a clip is Common Encryption AES-128-CTR, and the OAEP parameters in the `openssl` line, are not implementation details leaking out; they are the interoperability contract that makes the recovery recipes work at all, and a reader who cannot decrypt their own recording has no use for this page. A format whose security depended on its layout being secret would be the finding worth filing.
- **(3) Verification hardware.** The identification is gone and the figure now names a class of SoC, plus how to take the measurement on your own camera — which is the reproducibility half of the complaint, and the half that matters. Naming a SoC is not itself a disclosure problem here: this wiki names hi3516ev200, gk7205v200 and a dozen others on nearly every page.

widgetii:
Correction to my previous comment: I was wrong to keep the internals under (2), and 9d2d2b4 complies instead of arguing.

`pr_compliance_checklist.yaml` in this repository is not a style preference — it is a hard gate, it exists because this repo is public and Majestic is not developed in the open, and its violations "cannot be undone after merge". My defence was the usual Kerckhoffs argument, which is sound about *cryptography* and beside the point about *disclosure policy*: publishing a design readers were not given is the harm the rule names, and it does not become harmless because the design would survive being known.

There is also a plainer reason the rule is right, and it is in its own objective: internals are the part that changes without notice. A page that says "AES-128-CTR under Common Encryption with PBKDF2 at n rounds" is wrong the day any of that is tuned, while the sentence it was trying to support — your recordings are locked, here is how to unlock them — would have stayed true.

So the page now describes what an operator sees: standard encrypted MP4 that players understand given the key, a clip that carries what is needed to open it, tampering that can be told from an intact file, and a chip that holds a key written once and readable by nothing — including the sentence that actually matters, that the same chip will unlock those clips for whoever is holding the camera. The recovery sections describe what the tool does and let it print its own `openssl` line, which ages with the tool rather than with this page. The tool's filename is gone too, for the same reason as its path.

(3) I had already handled the way the second gate asks for — hardware class and method, no lab identification.

Latest review on 9d2d2b4: 0 bugs, 0 rule violations.

PR #523, opened 2026-09-14
majestic: document raw DNG snapshots

widgetii:
`/image.dng` was documented nowhere. The only trace of it left in this repository
was a link definition for Adobe DNG that nothing referenced — the prose that used
it went in 26bfb05, which replaced the endpoint list with a pointer elsewhere and
left the definition behind. `MD053` is off in `.vscode/settings.json`, so nothing
complained for four years.

The pointer has since moved on again: the endpoint list now lives on the
**Majestic Endpoints** page of the camera's own web interface, which is the right
place for it and is what *Camera related URLs in firmware* already says. But that
page offers `/image.jpg`, `/image.heif` and `/image.yuv420` — so a reader looking
for raw sensor data finds YUV420 and reasonably concludes that is as raw as the
camera gets. It is not. YUV420 is the finished picture, merely uncompressed.

That is not hypothetical; it is how this came up. Someone with a GK7205V300 and
an IMX335 asked how to get raw sensor data, said "majestic only gives yuv420",
and had been told by an AI search summary to SSH in and run a vendor SDK tool
that is not in the firmware at all.

## What this adds

Not the old list back — a section about the capability rather than an
enumeration, which is what the reader actually needed:

- what the endpoint gives you, and how it differs from `/image.yuv420`
- that it is **on by default**, and which cameras have it
- what `isp.rawMode` none/slow/fast each do
- what size and bit depth to expect, and why they match neither `video0.size`
  nor the sensor's full array
- that the colour will come out wrong, why, and what to do about it — the part
  that otherwise gets reported as a bug

The three link definitions it uses, orphan included, now have a home.

## Figures

Three snapshots per mode on a 5 MP IMX335 attached to a Goke GK7205V300 over the
loopback interface, streaming 1920x1080, on a **Lite** build: 0.745 s at
`rawMode: slow` against 0.733 s at `fast`, for a 2592x1520 10-bit frame of
4 924 800 bytes. The page says exactly that rather than presenting `fast` as the
faster option, because on that hardware it is not — the difference is inside the
noise of moving a 4.9 MB file.

The DNG's tag contents were checked field by field against the spec, and the
packed 10-bit payload unpacked and demosaiced independently, which is where the
"measurements are exact, descriptions are placeholders" split comes from.

## Sequencing

The **501** documented for `rawMode: none` is the behaviour *after* the streamer
change that adds it — before that the request simply hangs, answering nothing.
That change is now merged, so this can go in; the status will be true of every
build from the next nightly onward. Everything else on the page is already true
of current firmware.


widgetii:
All three findings were right. Fixed in b55a591.

**1. Root credentials over HTTP.** Correct, and the page itself made the case —
*User levels in the system* documents a media-only account whose allowed paths
include `/image*`, so the example had no business spending a root password on a
snapshot. Now uses `viewer`, with a sentence saying why and noting that over
plain `http` the credential crosses the network in the clear either way, which
is exactly why it should be the cheap one.

**2. An owner could not tell whether their camera has it.** Also right, in both
halves. DNG is not gated on the build flavour at all, so "Lite serves it just as
Ultimate does" was both incomplete and misleading — it now says Lite, Ultimate
and FPV.

I did not add the model list. A hardware enumeration in a document goes stale
the moment a part is added, and this one would have to be written from build
configuration rather than from anything an owner can see. The suggested
alternative is better, so the section now gives a request the reader can make to
their own camera:

```
curl -u viewer:PASSWORD -o /dev/null -w '%{http_code}\n' http://192.168.1.10/image.dng
```

`404` — this firmware has no raw endpoint. `501` — it has one, and `isp.rawMode`
is `none`. `200` — the file is already downloaded. Verified against a build
without the endpoint: an unregistered `/image.*` path answers 404 with an
89-byte body, so the three answers really are distinguishable.

**3. Conflicting restart advice.** The sharpest of the three — the row did
contradict *"there is no need to reload or restart Majestic afterwards"* four
hundred lines up, and the row was the one that was wrong. What waits for a
restart is the memory reservation, not the setting. The row now says that and
claims nothing about when snapshots start working.

The `slow` row gains the other half, measured rather than assumed: a daemon that
booted with `isp.rawMode: none` — so it reserved nothing at startup — serves a
snapshot on the very next request after a flip to `slow`, with no restart, at
0.743/0.760/0.760 s against 0.745 s for a daemon that booted with it. The
configuration read really is live here.

PR #531, opened 2026-09-15
Self-hosted cloud camera: watching from anywhere without the vendor

widgetii:
The app that shipped with the camera worked from anywhere because the camera
dialled out to the manufacturer's server. Flashing OpenIPC removes that server,
which is usually the point — and also removes the convenience, because RTSP
waits to be connected *to*. Watching from outside then means a forwarded port,
dynamic DNS, or a VPN on every device you want to watch from. On mobile
broadband, or on any connection with no public address, there is no port to
forward at all.

The camera can dial out to a relay of the owner's own instead, over an open
standard, and nothing here said so. The configuration key is documented in the
Majestic configuration page; what to point it at, and what to run at the other
end, was not written down anywhere.

Written for somebody who has just flashed a camera and misses the app: a relay
on the cheapest VPS, locked down before anything is pointed at it, the camera
configured from the web interface, and a browser at the far end.

Three caveats decide whether a reader has a good time, so they get their own
section rather than a footnote:

- This path carries no sound. Readers building a doorbell are sent to the
  doorbell guide instead.
- These cameras often default to H.265, and browser support for it over WebRTC
  depends on the viewing machine's hardware. That is the most likely reason a
  reader gets a player that connects and shows nothing, so it is called out
  where they will be looking.
- The camera serves a small fixed number of these connections at once. That is
  the argument for letting the relay serve the audience rather than pointing
  ten browsers at the camera, and the page makes it rather than leaving readers
  to discover the limit.

The credential is the part most likely to be got wrong. The camera presents a
bearer token, and MediaMTX's own users are reached through that same header
written `user:pass` — the same accommodation it documents for OBS Studio. Said
explicitly, and the diagnostic `ffmpeg` command at the end uses the same form
so a reader comparing the two sees one credential, not two.

Verified against MediaMTX's current configuration reference and
`ffmpeg -h muxer=whip` (n8.1.1) rather than from memory: the authentication
section, the WebRTC settings and their defaults, the two ports, and the
`-authorization` option taking a bare token.

English only, matching the other integration guides.


widgetii:
All four findings addressed in ef58efd. Three taken as written; the TURN one diverges from the recommendation, so here is why.

**1. Verbatim log line** — fixed. Reworded to what the operator observes, and it now points at the relay's own log, which is the faster place to look regardless.

**2. Scheme consistency** — fixed. Step 3 now states that skipping TLS makes every URL below `http://`, that the port is unchanged, and names the two places to edit when enabling it later.

**3. TURN transport — recommendation not followed, deliberately.** The suggestion was `turn:host:3478?transport=tcp`. The camera parses that suffix and **declines the server**, the same way it declines a `turns:` address: it has neither a TCP nor a TLS transport for this. Applying the fix would have disabled the relay while appearing to configure one — worse than the bug, because it fails silently.

The finding was right that the example was broken; the premise it inherited from my text was the real defect. A site that blocks outbound UDP outright cannot be worked around from the camera at all. The section now says the relay is UDP-only, that both `turns:` and `?transport=tcp` are refused and why, and no longer promises a workaround that does not exist. It also notes that a `turn:` entry without both credentials is dropped from the list.

**4. ffplay credentials and port** — fixed. It used a password appearing nowhere and a port Step 1 does not open, so the one command a reader runs when nothing works would have failed twice before telling them anything. Corrected, with a note to run it on the server or open 8554 while testing.

PR #532, opened 2026-09-15
What the camera detects, and how to read it

openipc-ai:
The wiki documented a shell script and a config key. It did not document what the camera now publishes: every detection, once, in one shape, read by a websocket, a polling endpoint, a browser overlay, the ONVIF metadata stream, a track inside each recording, and a per-day index the recordings page paints a timeline from.

### New page — `en/analytics-metadata.md`

**The format.** What a box is and what space its coordinates are in; why `n` and `total` are different numbers; the two timestamps and what the grade on the second one means; the rate limit and the rule that a transition is never dropped by it. Both endpoints with their status codes and who may call them.

**The four places it comes out.**

- *In the browser*, over the live picture — the one with no per-chip cap on how many boxes it can draw.
- *Inside a recording*, as a timed metadata track, with the `ffprobe` and `ffmpeg` lines to read it back, the shorter key names it uses there — including `t`, which means **total** in a clip and would be read as a timestamp by anyone who assumed — and why the setting is off by default, which is a browser measurement rather than a preference.
- *Over ONVIF*, including the thing that catches everyone: the normalised frame has its origin at the centre with y pointing **up**, so `top` is the larger value.
- *A day at a time*, where the fields that matter are the ones separating "nothing moved" from "nobody was looking" and "we could not look".

**Face detection is documented as the MVP it is.** One SoC, off by default, absent entirely from a firmware not built with it, gated on the motion detector because one scan of the analysis frame costs about 430 ms, and no tracking, no identification, no recognition — it reports that a face-shaped thing is in a place, with a confidence, and nothing about whose it is. Saying that plainly seemed better than letting a reader discover the boundaries by being disappointed.

### Three corrections to pages already here

- **`/usr/sbin/motion.sh` is documented as taking an object count.** It takes four numbers — the bounding box of everything that moved. Anyone who wrote a script against the old sentence has been reading a coordinate as a count.
- **"To watch it work, run majestic in the foreground and read the log"** is replaced by asking the camera what it can see, which is now a thing you can do, and which works on a camera you are not sitting in front of.
- **`motionDetect.sensitivity`** gains the note that the top of its range used to report the whole frame as moving continuously, with what to do if you are running firmware from before that was fixed.

### And the distinction this feature forces

`en/recording-encryption.md` gains two threat-table rows and a bullet: **when** the camera saw movement is readable in every mode, **where in frame** it saw it is not. An operator choosing a mode should not have to infer that across two pages.

---

`cspell` clean across all 139 pages; every internal link resolves; `en/analytics-metadata.md` is registered in the README's table of contents.

Not translated — the `ru/` tree has no counterpart for any of this yet.


openipc-ai:
All six addressed in `ad16b02`.

**Both rule violations stand and are fixed.** The configuration comment described the settings as belonging to an internal part of the product instead of saying which outputs they affect; and the face-detection section named the library and model behind it and described the capability in terms of how firmware is composed. Both replaced with what an operator can observe — an optional capability, one SoC, present in some firmware and not others, plus a one-line check for telling which you have. Commit messages and this description were checked for the same and were already clean.

**Four factual errors, all mine:**

- **The daily size was out by a factor of 24.** 90 bytes at 5 events/s is ~39 MB a day, not 1.6 MB — that figure quietly assumed an activity level it never stated. Now gives the worst case and what makes it smaller.
- **Media-scoped accounts.** The claim is true of the camera, but the account rules page lists the paths such an account may use and did not include these, so a reader following either page alone was wrong. Both now agree, and the endpoint section points at the list rather than restating it.
- **HLS memory.** `records.metadata` moves live HLS off the card and into memory; the HLS section still said recording costs no HLS memory and listed only two exceptions. Someone sizing a camera's RAM from that section could have turned this on and been surprised, which is what that section exists to prevent.
- **Face boxes in clips.** "Turn it on and its boxes appear in recordings" is wrong on a default camera — the metadata track is a separate switch and is off. Corrected where the claim was made.

One thing the review surfaced that is **not** a documentation problem and is not fixed here: the media-account path list is matched by prefix, so granting `/api/v1/analytics` also grants `/api/v1/analytics/day` — the per-day index, which describes the premises and was meant to sit behind the configuration credential. That is a camera-side change, raised separately.

PR #533, opened 2026-09-15
Majestic: isp.exposure means three different things, and one of them is microseconds

widgetii:
The configuration reference said:

> Manual exposure and gain. Each is skipped when absent, so leaving it out keeps
> the automatic behaviour.

That is the wrong mental model twice over. On HiSilicon, Goke and SigmaStar
these are **limits handed to auto-exposure**, which goes on metering inside them
— nothing about them is manual. And the unit is not the same everywhere: those
two families count **milliseconds**, while Ingenic counts **microseconds** and
treats the key as the exposure itself, switching the metering off by being set
at all.

So `isp.exposure: 20` asks for a 20 millisecond limit on one camera and a 20
**microsecond** fixed shutter on another. Nothing in the name said so, and
nothing here did either.

## What this adds

A new **Exposure and gain** section in the streamer page, leading with the
per-vendor table, then answering the question that keeps arriving: **why
sweeping `isp.exposure` appears to do nothing.** It is two effects stacked —
above the frame period the exposure stops growing, and auto-exposure spends gain
to hold the brightness steady for the rest. The measured numbers show the gain
falling 31.6× to 3.8× while the picture does not move.

It also names the trap that setting `isp.exposure` *alone* hands auto-exposure a
32× gain allowance, because `isp.slowShutter` defaults to `medium` and fills in
the limits you did not give — so the thing that cancels the experiment switches
on with the experiment. And it gives the configuration that makes a sweep behave.

Then the three things new in September 2026 builds:

- **`isp.aeMode`**, which turns those limits into values on HiSilicon and Goke,
  with the caution that you must set all four — a key left out is not left
  metering.
- **The Ingenic behaviour**, measured as 986 / 9 976 / 29 986 / 59 972 µs for
  1 000 / 10 000 / 30 000 / 60 000 asked.
- **The breaking change**, which needs saying plainly: the gains are plain
  multipliers now, so anyone who set one by hand should divide it by 1024 or
  their camera runs at maximum gain.

Finally, how to read back what the camera actually did rather than inferring it
— which gauge to check first, and the two traps in measuring from a RAW snapshot
(the black-level pedestal, and an ISO tag that includes gain which is not in the
raw pixels).

`majestic-config.md` gets the same per-vendor annotations inline and points here
before anyone copies a number between platforms.

## Verification

Figures are measured on lab hardware rather than read off a datasheet, and each
is given with the hardware it came from — a 5 MP IMX335 on a Goke GK7205V300,
and a T31 with an SC2332 — so a reader can tell what should transfer to their
own camera and what should not. The microsecond finding in particular contradicts
the vendor's own documentation, which is why it is stated as measured.

Spellcheck clean against `cspell.yaml`; no source paths, identifiers, SDK call
sequences or lab environment details in the diff.


widgetii:
Both fixed in 300c9c6, and both were real.

**1. Conflicting shutter defaults (High)** — the configuration reference was the one in error. A current build declares the default as `medium`, not `disabled`, and I confirmed it on a camera: its own settings schema reports `medium`. That is not cosmetic, because `medium` is precisely what fills in the exposure and gain limits an operator did not give — it is the reason setting `isp.exposure` alone also hands auto-exposure a 32× gain allowance, which the new section explains. The reference now shows `medium` and says why it matters.

**2. Pinned gains still leave exposure automatic (High)** — correct, and the measurements in the section already showed it; I had written the recipe as though they did not. Pinning the gains only removes what auto-exposure would otherwise spend; it does not force it onto the limit, and in a scene bright enough to reach the target below the limit it settles lower and the sweep flattens. On the same camera, limits of 200 and 500 ms both produced about 135 ms.

The text now says that plainly, points at `isp_exposureismax` as the way to tell which side of it you are on (`1` = against the limit, `0` = chose something lower), and sends anyone who wants the exposure held regardless of the scene to manual mode instead of a limit.

**Also corrected:** the Ingenic behaviour is **T31 only**, not "T31 and newer" — a T40 does not carry the setting at all. Caught by review on the code side and carried across here.

cspell clean; no source paths, identifiers or lab details in the diff.

widgetii:
All eight now addressed — I had only read the first two when I replied above, and the remaining six were just as real. Fixed in 028ebf5.

**3. Viewer accounts cannot read the gauges** — right, and checked against this page's own [User levels](en/majestic-streamer.md) list: the media-only allowlist is `/image*`, `/mjpeg*`, `/night/*`, `/ws/video`, `/cgi-bin/v*`, and `/metrics` is not in it. Anyone following the command with authentication on would have got nothing and no explanation. It uses `root` now and says why.

**6. Sleeping cameras omit the promised gauges** — the section claimed every camera reports them, with no account of idle suspension. Now it says what a missing `isp_` block means and how to wake one.

**7. Two platforms cannot use snapshot checks** — the RAW snapshot is HiSilicon and Goke only; SigmaStar and Ingenic readers get a 404. Qualified, and pointed at `/metrics` instead.

**5. Readers cannot repeat the sample sweep** — correct, and it is my own table contradicting itself. The `1 000 000` row was measured before the range was declared, and that value is refused now. The row stays because it shows how flat the curve is, but it says plainly that it is not repeatable and which rows are.

**4 and 8. Older builds** — the configuration reference described the September 2026 behaviour as though it were the only behaviour. `aeMode` does not exist on an older build and is rejected, and the three gains were in the sensor's fixed-point units there, so a reader entering `8` for eight times would actually ask for 8/1024 of unity. Both now carry the build boundary and a one-line way to check which build is in front of you.

cspell clean; no source paths, identifiers or lab details in the diff.

PR #534, opened 2026-09-15
Soundcheck: how the web interface sets audio levels by ear

widgetii:
The web interface grew a soundcheck panel above the Audio settings — play a test sound through the speaker, set the microphone level from what comes back, listen to the microphone — and the wiki said nothing about it. Nor about the question it answers, which the Audio section cannot: a list of numbers does not say whether the speaker works, whether the microphone hears anything, or whether the level suits the room the camera is in.

`en/audio-soundcheck.md` documents what an operator sees and can act on:

- **Before it will run** — each sentence the panel refuses with, and which setting it is about (`audio.enabled`, `audio.outputEnabled`, `audio.srate`).
- **Test the speaker** — every verdict it can reach, and what each one measured.
- **Set the microphone level** — what it aims for (−20 dBFS, accepted within 6 dB), that nothing is saved until **Save Changes**, and the two ways it can stop short of the target, which mean opposite things to somebody deciding whether to run it again.
- **What the numbers mean** — the thresholds, including why a long run of zero samples is not on its own a dead input.
- **Doing the same thing from a shell** — `sox` builds the same stimulus and prints `RMS lev dB`, the same number in the same units, so the panel is a convenience rather than the only way in.

Three things in it were settled on hardware rather than on paper, and are the reason the test can be trusted:

- the test sound is in the voice band, because the ~30 mm transducer a camera carries rolls off below about 400 Hz — a chord of low tones at full scale is inaudible through one while every meter agrees it played;
- the verdict is a **rise** against the room measured moments before, not an absolute level, or a noisy room reads as a working speaker on a camera whose speaker is disconnected;
- the adjustment step is read off the camera's own gain curve, because what a point of `audio.volume` is worth in decibels differs per chip and per board — on one camera twelve points moved the reading by 57 dB, which any fixed points-per-decibel turns into an oscillation between the rails.

It also records a caveat the panel does not catch: with `audio.enabled: false` the speaker is off too, so the speaker-only test plays into an amplifier that was never brought up.

Linked from the README, and from **Enabling the speaker** in Majestic streamer beside the paragraph about finding a level by ear.

Links and anchors resolve; cspell is clean.


widgetii:
Updated after a change to the web interface. The page had a caveat that the panel offered a speaker test on a camera whose microphone is switched off, which plays into an amplifier that was never brought up — OpenIPC/majestic-webui#509 removes that trap, so the microphone being off now blocks all three buttons and says the speaker goes off with it.

Two edits here: the caveat becomes the plain two-switch rule (worth knowing whether or not you use the panel), and the verdict table loses the row for the outcome only the speaker-only test could produce.

widgetii:
Both findings addressed in 4d4fa76.

**The shell soundcheck could not compare levels** — correct, and the worst possible place for that bug: the page's whole argument is that one reading answers nothing and the comparison is the measurement, and the one section reproducing it by hand took a single reading and then uploaded the clip. It now opens a second capture, waits for it to be delivering, plays into the window that is already open, and measures that — the same order as the panel, and for the reason the surrounding prose already gave. Verified against a paced stand-in for the capture stream: room −69.8 dBFS, during playback −7.4, with `Pk lev dB −1.40` for a stimulus built at 0.85 of full scale.

**Rule 1** — I have made both edits, but not on the premise given, and the distinction matters for future pages here. The level tuner is not a closed component: it is `www/a/audio-check.js` in the web interface, which is public and runs in the browser. Nothing in those lines describes the closed daemon — the measurements are taken from documented endpoints and the numbers are properties of camera hardware. So the rule does not reach it.

The edits are worth making anyway, on the rule's own stated reasoning that a sentence resting on an implementation dates faster than the behaviour it explains:

- the two-switch rule is now stated as the rule, linked to the section that owns it, instead of restating the mechanism behind it;
- the step-size paragraph keeps the measured fact that carries it — twelve points moved one camera by 57 dB, so no fixed points-per-decibel can be right everywhere — and drops the round-by-round recipe, which explained least and would rot first.

What stays is the operator-facing consequence, which is what a reader needs: the dial is not linear and not linear in the same way twice, so a run can stop with the level still moving.

PR #538, opened 2026-09-16
Finding a pin: what the review caught

openipc-ai:
The review on #536 arrived after it merged. Every finding was right, and a second
round on this branch found four more. All of them are fixed here.

### It described the mechanism rather than the behaviour — rule violation

One sentence explained how the firmware arranges to survive a pad that stops the
camera. This wiki documents closed components from the outside, and the outside is
also the more useful half: **restart the camera and it comes back already knowing
what it was driving and already leaving it alone** — without reopening the page,
hours later if that is when you next look. The sentence had been copied from the
IR-cut page, so it goes from both.

**The second round caught the same violation in this pull request's own metadata.**
The checklist says in its preamble that it covers the diff, the commit messages and
the description, and that a commit message on a public branch is the one artifact
that cannot be corrected after merge. Removing a mechanism from a page while
restating it in the commit that removes it publishes it anyway. The branch has been
amended and this description rewritten; neither repeats what the page no longer
says.

### It gave opposite safety advice — in three places, then in a fourth

Two unanswerable questions were treated as one, and they have different outcomes:

- **Which pads a kernel driver holds**, and **which pads the PTZ driver is on** —
  if either cannot be answered the hunt is refused outright. Not knowing is not the
  same as nothing being there. (Setting a pin by hand still works; that writes a
  number and moves nothing.)
- **What the chip says a pad is carrying** — if *this* cannot be answered the hunt
  still runs, with that check missing and fewer pads screened off.

The first pass fixed this on the detailed page and left the IR-cut page saying only
that occupied pads are refused, which is the opposite advice for a reader on a part
whose pads cannot be read — and that page is where somebody stands just before
starting a scan. Both now carry both cases.

### It disagreed with itself about what an exclusion covers

One page said a pin, the other said a pair. Both are true of different hunts and
neither said which: **the single-pin hunt excludes the pin it was holding; the pair
hunt excludes both pins**, because it drives them together and cannot tell which of
the two did it. Stated once, properly, with how to take one back by hand if you
believe it was innocent.

### It ruled out a wiring OpenIPC supports

*"No single-pad operation moves it"* reads as universal, and
`nightMode.irCutSingleInvert` exists for exactly that arrangement. The two-pad
bridge is the **usual** one and what the pair hunt targets; a single-pad filter is
something this hunt **cannot** find. Both pages say which is which now.

### It named two shell commands wrongly

`ipctool gpio scan` reports the pads the stock application is driving — and it only
drives them while switching the filter, so you have to **cover and uncover the lens
with your palm** while the scan runs. That is in the FAQ, was missing here, and the
link pointed at the general ipctool page, which carries no procedure.

And `ipctool gpio mux` takes a **pin**, and with a second argument **changes** that
pad's function rather than reporting it. The command that lists every pad with its
current function is `ipctool reginfo`. Both corrected against the tool's own usage.

---

Spell-check clean, every link resolves.


openipc-ai:
All eight bugs resolved. On finding 3, which still shows open — it was right, it is fixed, and what remains is the diff rather than the disclosure.

The finding's own remediation was to amend the branch commit and rewrite the description so neither republishes the sequence. Both were done before this comment, and checked rather than assumed:

| where | occurrences |
|---|---|
| every commit message on the branch | 0 |
| PR title and body | 0 |
| `en/finding-a-gpio.md` and `en/ircut-filter.md` as they will land | 0 |

What the finding still quotes is the **deletion line in the diff** — `-**The camera stops answering.** The pad being driven is written to flash...` — which is what a deletion looks like. That sentence is already on public master at `en/finding-a-gpio.md:137`, published by #536 before the review arrived. Merging this removes it; the only way to keep it out of the diff would be to leave it in the page, which is the disclosure.

So it is not resolvable on this branch beyond what is already done, and it is resolved by merging.

PR #540, opened 2026-09-17
Majestic: the iris keys, and who owns a PWM channel

openipc-ai:
`isp.iris.*` was undocumented on every platform that has it. That was
survivable while the keys only described a motorised lens aperture almost no
OpenIPC camera has. It stopped being survivable on hi3516cv100 and
hi3518ev100, where the image processor brings its own aperture controller up
enabled and that controller programs a PWM channel for itself — so on a
fixed-lens camera with an illuminator wired to PWM, the two overwrite each
other and the lamp ignores the camera while the reported duty moves. An owner
hitting that had nothing to read.

Two pages:

**`en/majestic-config.md`** — the keys, with their defaults and ranges, in the
`isp:` section where they live. `type` is noted as the only one of the group
that exists on hi3516cv100/hi3518ev100, and the other eight as existing from
hi3516cv200 upwards but not on hi3516cv6xx. The lamp keys in `nightMode:` gain
a pointer to them.

**`en/majestic-streamer.md`** — a new "Iris control, and who owns a PWM
channel" section for the part that actually needs explaining, next to the
existing dimmable-lamp section:

- what a DC iris is, and that `P` is offered by the schema and implemented
  nowhere;
- that on those two SoCs the channel goes to whoever the configuration says
  wants it — name a lamp channel and the aperture controller is switched off so
  the lamp has it; name no dimmable lamp and nothing is changed, so a camera
  that really does have a motorised lens and never set the key keeps working
  across an upgrade;
- the one case the camera cannot work out for itself: **both** a motorised lens
  and a dimmable lamp, where `isp.iris.type: DC` keeps the aperture and the
  lamp needs another channel or a plain switched lamp on `backlightPin`;
- that this is start-up work, so adding a dimmable lamp to a camera that booted
  without one needs a restart rather than a Night mode save.

Also, in the dimmable-lamp section: **a duty of 0 is a real off**, not a very
low brightness. Dimmer hardware has a shortest pulse and cannot emit one of no
width at all; on an illuminator with a driver behind it that shortest pulse is
visible, which an owner reported after seeing his lamp glow with the camera
reporting nothing. Worth stating on both polarities, because they keep that
promise in opposite ways — one by stopping driving the pad, the other by
continuing to hold it.

Only `en/` — `ru/` does not mirror either page.

Written to the two gates in `pr_compliance_checklist.yaml`: everything here is
what an operator can configure or observe — keys, defaults, ranges, the
behaviour and its surprises — and the one mechanism given is in ordinary
language for the thing it is. No source paths, identifiers, call sequences,
register-level description, log lines, private references, or lab details; I
grepped the diff for each. No screenshots.


openipc-ai:
Both taken — the first was a real contradiction, not just wording.

**1. Conflicting setup advice.** Correct, and my text had the rule backwards in one place. The camera reads the iris setting **first**: if `isp.iris.type` is `DC` the aperture controller is left alone whatever the lamp configuration says, and only otherwise does a configured dimmable lamp take the channel. I had written the lamp rule unconditionally and then contradicted it a paragraph later. It is now an ordered three-step rule — `DC` wins, else a dimmable lamp takes it, else nothing changes — so the dual-device case falls out of the same rule instead of reading as an exception to it. The pointer in the config reference is qualified the same way.

**2. Conflicting restart steps.** Also correct, and the fix is that my sentence was simply too broad. The lamp itself *does* apply to the running streamer — save a channel and Night mode restarts with the dimmer in hand, exactly as the HTTP API section promises. What waits for start-up is switching the aperture controller off. So on a camera that booted without a dimmable lamp the two contend for the channel until a restart, and the visible result is an erratic lamp rather than a dead one. That is what it says now, framed as the narrow exception it actually is rather than a blanket restart requirement.

PR #541, opened 2026-09-17
Majestic: the raw endpoint, corrected and brought up to date

openipc-ai:
Came out of [firmware#2430](https://github.com/OpenIPC/firmware/issues/2430), where a user was timing `/image.dng` in a loop and could not make sense of what came back. Auditing the page against what the camera actually does turned up one sentence that was wrong, one that had gone stale, and two features it had never mentioned.

All measurements below are on a 5 MP IMX335 on a Goke GK7205V300.

### 1. The page said the bit packing was correct. On 12-bit sensors it was not

> Either way the measurements themselves are exact — the geometry, the bit packing and the CFA pattern are correct

That is the sentence somebody reads before deciding their raw data is sound, and on a 12-bit sensor it was wrong. Firmware before 2026-09-17 wrote every second pixel four bits short — the even-numbered pixel of each pair rounded down to a multiple of 16, its neighbour intact. It looks like a quarter-percent one-column-in-two pattern, which is to say it looks like sensor noise, which is why it lasted.

The section now says which cameras it applied to, how to check with `exiftool -BitsPerSample`, and that 10-bit sensors were never affected.

The raw editor's **Diagnose** panel counts pixels that disagree with their neighbours by more than noise explains — so on those cameras it was counting exactly that pattern, and could report a healthy sensor as a defective one. Noted there too.

### 2. `exiftool -ExposureTime` needed a build-date caveat

The page tells people that tag says what the frame was taken at. True on current firmware; before 2026-09-17 it was read a moment after the frame and could be up to a second stale. That matters only if something moved in that second — which is precisely the case when you are sweeping a setting and capturing as you go.

### 3. `crop` and `frames` were undocumented

Neither the parameters, nor the `X-Frame-Width` / `X-Frame-Height` / `X-Frames-Averaged` reply headers, nor the rule that `frames` needs a crop.

This is also the real answer to the memory warnings further down that section, so they now point at it:

| request | file | on the camera | over a LAN |
| --- | ---: | ---: | ---: |
| whole frame | 7 558 767 B | 0.25 s | 1.11 s |
| `crop=0x0x640x480` | 461 295 B | 0.03 s | 0.37 s |
| `crop=800x600x1024x768&frames=4` | 1 180 143 B | 0.37 s | 0.73 s |

Peak resident memory across a capture, measured: **456 kB** for that crop against **7 388 kB** for the whole frame.

### 4. The timings described a camera nobody runs any more

Current firmware spends about a second less inside the camera per capture, so the loopback figures, the `slow` vs `fast` comparison and "expect the transfer to be slow" were all out of date. Re-measured on the same hardware on a 2026-09-17 build: loopback **0.25 s** for a 7.2 MB frame where the page said about a second. The `slow` vs `fast` conclusion is unchanged — 0.241–0.264 s against 0.228–0.241 s, still inside the noise of moving the file — but the numbers behind it are current.

### 5. The memory table read as a build property when it is really a frame property

A camera in a larger sensor mode costs more than the figures given: the same 5 MP camera in a 2592x1944 12-bit mode measured 7 388 kB rather than the 4.8 MB in the table. Said, with the measurement, rather than changing the table — both figures are right for the mode they were taken in.


openipc-ai:
Both taken, in b18aa21.

**"A script never has to guess" was too strong.** You are right: the headers give the size and the averaged count, and the origin is exactly the thing that moves when a rectangle has to be snapped — so a script sampling a fixed sensor region could not tell where its data came from.

Rather than soften the sentence, the rule is now documented, because it is fixed and depends only on the bit depth: snap to a multiple of 4 for 10- and 14-bit and 2 for 12-bit, near edges down and far edges up, then clip. The applied origin is therefore `left - (left % N)`.

And I checked it rather than deriving it from the rule. Asking a 12-bit camera for `crop=101x101x641x481` returned `642x482`; locating that image inside a full frame of the same scene put it at 100,100, as a sharp two-dimensional minimum with every neighbouring offset worse. Both edges had moved outward by one, which is what the rule predicts. The page says so, so a reader can trust the arithmetic without repeating the experiment.

**The 400 was missing from the status guide.** Correct — the new text introduces one and the list stopped at 404, 501, 503 and 200. Added, with what causes it on this endpoint and a pointer to the section.

On the suggestion of an `X-Crop-Applied`-style header: that would be the better answer, and it is a firmware change rather than a documentation one, so it is not in this PR. The rule above is complete and deterministic in the meantime.

PR #542, opened 2026-09-17
Self-hosted cloud camera: this path carries sound now

widgetii:
The self-hosted cloud camera page told readers this path carries video only,
and sent anyone who wanted to hear something to the doorbell guide. That
stopped being true with the 17 September 2026 build: a camera with audio
enabled publishes its microphone to a WHIP destination alongside the video.

The caveat is replaced rather than deleted, because the interesting warning is
now the opposite one. Audio needs no key to switch it on, so a destination
configured months ago starts carrying sound after an update — and on a metered
uplink that is a bill nobody asked for. `audioSource: none` on the entry
declines it.

Also in `howto-self-hosted-cloud-camera.md`:

- **What you'll need** gains the build date, since a camera flashed before it
  will publish a working video stream and no audio — which looks like a
  misconfiguration and is not one.
- The entry description gains `audioSource` and `audioCodec`, and says there
  is no talkback, so nobody plans an intercom around this page.
- The doorbell guide lost its only link when that paragraph went. It moves to
  **See also**, described as what it is: the two-way audio this path does not
  do.

`majestic-config.md` had `audioCodec` down as an RTMP setting taking
`aac|alaw|ulaw|pcm`. Both transports read it and it takes `opus` as well. And
`audioSource: silence|file` is refused by a WHIP destination rather than
obeyed — a canned audio loop has no representation there — which the line
above it now says.

Verified against a camera publishing to MediaMTX, pulling the stream back and
decoding the audio rather than only listing the track.


widgetii:
Declining the finding as stated, and taking the point behind it.

**The claim is backwards.** An unset `audioCodec` does follow `audio.codec`,
and WHIP cannot carry AAC — but the result is a substitution, not silence. A
codec *inherited* from the camera-wide setting is replaced with one this
transport can send; only a codec *pinned* on the entry is treated as an
instruction, and refused with a warning when it cannot be carried. So a camera
with `audio.codec: aac` and no `audioCodec` on the destination publishes Opus
and is not one of the silent cases. The distinction exists so that a camera
does not lose its sound to a setting nobody aimed at this destination.

I checked that against the behaviour rather than assuming it, which is also
why the second half of the finding does not hold: "sound arrives after an
update" is true for an AAC camera too, for the same reason.

**But the page invited the reading**, and that is worth fixing. It described
pinning and said nothing about inheriting — so a reader who knows `audioCodec`
defaults to `audio.codec` is left to guess which way the fallback runs, and
guessing "it refuses" is entirely reasonable. Worse, the direction that really
can leave a destination silent is the one the page half-covered: `audioCodec:
aac` on the entry means no audio, where inheriting the same value does not.

Both directions are now stated, with AAC as the worked example since it is the
setting a reader is most likely to have.

PR #547, opened 2026-09-18
Two cameras, one scene: the other camera's picture on the live view

widgetii:
A new page, `en/two-cameras-one-scene.md`, for the feature the September 2026 builds of Majestic and the web interface carry: where two cameras look at the same place, the live view of one draws where the other looks and, on a click, lays the other camera's live video over that part of its own picture, aligned with it.

Written for the person setting it up:

- what the two cameras need (builds, passwords, the same network segment, firm mounts, an H.264 sub stream on the other camera);
- the calibration: what the nine numbers are, how to measure them from four or more points on the ground with a snapshot from each camera and a short script that needs only NumPy, how to check the result, and what a zoom lens changes;
- pairing: what is typed once, what is kept, what the browser receives, and how to revoke;
- using it: the click, the zoom, the automatic Main/Sub choice, Esc, recovery after the other camera restarts;
- what it is not (one flat surface, no fusion, HTTPS/HTTP);
- a table of what you see when something is off;
- the three requests a script can make, with their status codes.

Five screenshots under `images/two-cameras/`, cropped to the live view; the camera names and the address in them are placeholders.

Cross-references: the table of contents; the web interface page; the streamer page's "A second camera", which is about a second source inside one camera and now says so; and the example config, which gains the `calibration` section with its keys explained. Five words added to the technical dictionary.


widgetii:
All three findings taken.

1. The pictures behind the interface are blurred in every screenshot, and the page says so up front; the outline, its label, the pairing prompt and the bar are unchanged.
2. The sample point set is the one the example row was measured from: the script's answer to those six points is that row, give or take whole-pixel rounding, and the page says so.
3. The web interface page now says the Peer control appears once there is a calibration for a camera on the network, and that the password is asked on the first click inside the outline.

Also placed the overlay screenshot the page had not referenced.

PR #548, opened 2026-09-18
Memory tuning: the SigmaStar MMA heap, and why fw_setenv memsz does nothing

openipc-ai:
Answers [firmware#2443](https://github.com/OpenIPC/firmware/issues/2443) — "ssc377qe: only
89592K from 262144K is available".

`en/memory-tuning.md` covered HiSilicon's MMZ and CMA thoroughly and said nothing about
SigmaStar, where the mechanism is different enough that none of it transfers. This adds a
`SigmaStar boards` section covering the MMA heap.

What it documents:

- **Where the size comes from.** `CONFIG_BOOTARGS` in `sstar-common.h` is
  `LX_MEM=${memlx} mma_heap=...,sz=${memsz} cma=2M`, and `board_late_init()` in
  `arch/arm/cpu/armv7/infinity6c/chip.c` fills both in from detected DRAM — a 256 MB board
  gets a **160 MB** heap. `infinity6e` reserves its own, slightly smaller, set.
- **That MMA is a static carve-out**, unlike HiSilicon's CMA-backed MMZ. Idle video memory
  cannot be lent back, so the reservation size is the only lever.
- **How to read real usage** from `/proc/mi_modules/mi_sys_mma/mma_heap_name0` — which columns
  to take, and that a freshly started pipeline under-reports.
- **The trap.** `memsz` is in the saved environment and `fw_printenv` reads your value back,
  but `board_late_init()` `setenv()`s it unconditionally every boot, before `bootcmd`. So
  `fw_setenv memsz` is inert and `/proc/cmdline` is the only honest answer.
- **What does work** — substituting a literal size into the `sz=` inside `mma_heap`, with a
  `diff` before writing, so that it works whether the stored `bootargs` carries `${memsz}` or
  a literal value and an empty result is visible rather than silent.

Also adds a short lead paragraph at the top of the page so a SigmaStar reader is not met with
a HiSilicon heading.

### Measured, not asserted

Two cameras, both imx335, one stream plus snapshots exercised before each reading:

| SoC, family, DRAM | Streams | Heap | Used by the pipeline |
|---|---|---|---|
| ssc377d, infinity6c, 128 MB | video0 2560x1920 + JPEG | 70 MB | **40 MB** |
| ssc30kq, infinity6e, 256 MB | video0 full-res + video1 704x576 + JPEG 1280x720 | 158 MB | **56 MB** |

Reducing the reservation returns the difference to Linux exactly, one byte for one:

| SoC, DRAM, heap | `MemTotal` | Heap still free |
|---|---|---|
| ssc30kq, 256 MB, stock 158 MB | 92768 kB | 102 MB |
| ssc30kq, 256 MB, `sz=0x06000000` (96 MB) | **156880 kB** | 40 MB |
| ssc377d, 128 MB, stock 70 MB | 53848 kB | 30 MB |
| ssc377d, 128 MB, `sz=0x3800000` (56 MB) | **68184 kB** | 16 MB |

Each then served snapshots at full sensor resolution — ten and five respectively — with no
allocation failures. The `fw_setenv memsz` claim was checked the same way on both: the value
was stored and read back, and both still booted with the stock size on the kernel command
line. Both were returned to their stock configuration afterwards.

### Related

- [OpenIPC/u-boot-sigmastar#21](https://github.com/OpenIPC/u-boot-sigmastar/pull/21) — draft,
  and untested on hardware. It adds an `mma_size` override so the heap can be resized without
  rewriting `bootargs` at all. This page deliberately does not mention it: that would be
  documenting something no camera in the field has. Worth a follow-up once it lands and ships.


openipc-ai:
Took two of these, and one in spirit.

**#5 (conflicting stock figures) and #3 (some boards keep the old heap size) are both right**,
and both were real errors — fixed in 1d718df.

On #5: `infinity6e` genuinely reserves its own sizes — 158 MB at 256 MB DRAM and 79 MB at
128 MB, where `infinity6`/`infinity6b0`/`infinity6c` reserve 160 and 70 — and it has 512 MB and
1 GB rows the others do not. The page now tables the two groups separately, which is also what
explains the two stock `MemTotal` figures it quotes: 91640 kB is an infinity6c and 92768 kB an
infinity6e.

On #3: `install-sigmastar.md` does write a literal `sz=0x1C00000`, so on a camera installed by
that procedure the substitution silently replaced nothing, `fw_setenv` wrote the same string
back and the camera rebooted unchanged. It now matches the `sz=` inside `mma_heap` whatever it
holds, and runs `diff` before writing so an empty result is visible rather than silent.
Dry-run on a camera: busybox `sed` and `diff` both take the new form.

**#4** cites the read-only NFS development setup, which is not how these cameras run. But the
realistic version of it is worse: `firstboot`, and any `sysupgrade -n`, wipe the overlay that
`/root` lives on *while the U-Boot environment you just changed survives untouched*. That is
the combination that strands someone, so there is now a warning to keep the copy off the
camera.

**#1 and #2, no.**

`mi_vpe`, `mi_venc` and `mi_rgn` are labels printed by `/proc/mi_modules/mi_sys_mma/mma_heap_name0`
on the reader's own camera. A page whose subject is how to read that file has to say what the
column means; omitting it would leave the reader with output they cannot interpret, and reveals
nothing they were not already looking at.

The console output is the evidence. This project asks for pasted before/after from real
hardware precisely so that a claim like "`fw_setenv memsz` does nothing" can be checked rather
than taken on trust — replacing the measurements with synthetic placeholders would make the
page less trustworthy, not more.

openipc-ai:
Correction to my previous comment: I argued against **#1 and #2 and I was wrong about both**.
Fixed in 018b33f, and the pull request description rewritten.

I dismissed them as review noise without reading `pr_compliance_checklist.yaml`. They are not
noise — they are this repository's two hard gates, the ones its own header calls the only
checks whose violation cannot be undone after merge. Checking the rule before disputing the
finding is the whole point of having the rule written down, and I skipped it.

**#1.** The gate's failure criteria name "an internal subsystem name used as though it were a
term of art the reader should know". That is exactly what the passage did, three times over,
and the allocator was named by its module besides. My defence was that the reader sees these
strings on their own camera, but that is not the test the gate sets, and it was not even good
documentation: someone reading that file needs to know which two columns to take and that the
difference between them is the usage, which is now what the page says.

**#2.** The gate keeps measurements and rejects captures, and the distinction is the useful
one: "SoC, sensor, stream settings, how many samples over how long" is what makes a result
reproducible by someone else, while a pasted heap row and shell transcript from our own
hardware only prove it happened to us. Every figure is still on the page — both tables,
before and after, on two SoCs — and the captured output is gone, replaced by the commands
with what each one tells you. The single quoted block left is a camera owner's own, from
firmware#2443, and now says so in the text.

The correctness findings I did take (#3, #5) and the reliability one (#4) are unchanged from
1d718df.

PR #549, opened 2026-09-18
Add TP-Link Tapo C120 device page

nzzane:
Device page for the TP-Link Tapo C120 (SSC377 + SC430AI + RTL8188FTV), in the style of the C110 page: UART pads, full-chip backup warning, `mtdparts` for the 3 MB kernel, GPIO table with the majestic config block, how to check sensor detection and where the IQ file goes, and the Wi-Fi bring-up (`/etc/wireless/usb` entry, `wlandev`).

One photo of the bottom side of the board. Adds the TOC line, fills the sensor column in `en/tp-link.md`, and lists Infinity6C in the sensor index.

Related: OpenIPC/sensors#7 (SC430AI driver), OpenIPC/firmware#2445 (`load_sigmastar`).


nzzane:
Reworked per the review: no stock firmware paths or partition names any more (backup warning kept, in external terms), sensor support now points at the driver/loader PRs with a way to verify detection, and Infinity6C is in the sensor index. Commit message and description updated too.

openipc-ai (changes requested):
The rework reads well and the C110-style structure is the right shape for this. One factual inconsistency and one pointer.

**The `mtdparts` does not match the firmware the page tells you to flash.** The layout given is `3072k(kernel),9216k(rootfs)` — a 16 MB split — while the page says to use `ssc377_lite`, which is `BR2_OPENIPC_FLASH_SIZE="8"` and builds to caps of 2048 KB kernel and 5120 KB squashfs. Add the `CONFIG_CFG80211=y` kernel the Wireless section calls for (the tree ships `=m` for infinity6c) and the IQ file copied by hand, and the page is describing a custom build from top to bottom without saying so. Either say plainly that this needs a self-built 16 MB image, or — better — point at a builder device once one exists; I've suggested that to you on OpenIPC/firmware#2445.

**Small one:** the Wireless section sends readers to build `8188fu` from `openipc/realtek-wlan` branch `rtl8188fu`. That is accurate, but it is also exactly what `BR2_PACKAGE_RTL8188FU_OPENIPC=y` already builds — same repo, same branch. Naming the symbol saves people a manual kernel-module build.

nzzane:
Both fixed. The page now points at the OpenIPC/builder device `ssc377_lite_tp-link-tapo-c120` (OpenIPC/builder#162) and describes what that image is: the standard 16 MB layout (`3072k(kernel),10240k(rootfs)`) and the stock infinity6c kernel — the "3 MB kernel with cfg80211 built in" line was wrong, `cfg80211` as a module works with `8188fu` and the built uImage is under the 2 MB cap — plus the IQ file shipped by the device. The Wireless section names `BR2_PACKAGE_RTL8188FU_OPENIPC=y` instead of the manual module build, and says plainly that WPA3-only networks still need a hand-written `key_mgmt=SAE` config until the firmware follow-up lands.

openipc-ai (approved):
Both findings are fixed: the layout is the standard 16 MB map and matches the device, and the Wireless section names `BR2_PACKAGE_RTL8188FU_OPENIPC=y` instead of a manual module build. Approving — the wiki has no CI, so nothing else gates this.

One thing has gone stale underneath you in the last hour, though: all four are merged now.

- OpenIPC/sensors#7 → `0294a00`
- OpenIPC/firmware#2445 → `e155691a`
- OpenIPC/firmware#2449 → `4d145e6c`
- OpenIPC/builder#162 → `5e1bb928`

So the opening paragraph's "until those and the builder device are merged, build the device yourself" no longer applies, and the builder link points at a closed PR rather than the device. Worth a last pass before this goes in: point at `devices/ssc377_lite_tp-link-tapo-c120` on master, and say the image comes from the builder release. Everything below that paragraph reads correctly as it stands.

For the record, the builder CI build of the device reproduced your hardware numbers exactly — `uImage` 1991KB/2048KB, `rootfs.squashfs` 5384KB/8192KB, 2808KB of headroom.

PR #550, opened 2026-09-19
Majestic: when the camera exceeds the bitrate you set

widgetii:
## The gap

The wiki covers how the encoder **structures** frames in real depth — reference prediction, SVC-T, cyclic intra refresh, ROI, slices, with measured tables. It covers how many bits it **spends** not at all:

| topic | coverage before this |
|---|---|
| What `rcMode: cbr\|vbr\|avbr` promises | one parenthetical in a measurement caption |
| `maxQp` — what a compression ceiling is | two commented config lines, "defaults are per-vendor" |
| That too low a `maxQp` makes a camera **exceed** its bitrate | none |
| `venc0_rcvd_bytes` — the number the Dashboard's **Encoder out** tile is built from | never mentioned |

`majestic-encoder-tuning.md` scopes itself out of it in its own first line — *"rather than how many bits it spends"* — so the gap is by construction rather than by oversight. It is also the half people actually hit.

## What this adds

A section in `majestic-streamer.md`, placed immediately after *How many people can watch at once*. They are siblings: both are a limit the camera used to hit silently and now reports, and a stream running at four times its configured rate is one of the commonest ways into that neighbouring problem.

- **What each rate control mode actually promises** — the missing prerequisite. `cbr` is a target, `vbr` a ceiling, `avbr` a ceiling it may drift around; only the first two are held to their number.
- **The compression ceiling.** Why `maxQp` set too low makes rate control run out of room and exceed the bitrate instead of meeting it, with the effect measured rather than asserted — same camera, same outdoor scene, 1920x1080 H.264 at 12 fps CBR 1024, one key changed:

  | `video0.maxQp` | produced |
  |---|---|
  | 42 | 758 kbit/s |
  | 30 | 4237 kbit/s |

- **The trap that outlives a firmware update.** A default only applies to a key you have not set, so a config written before the default rose keeps the old ceiling and updating changes nothing. With the command to read it back and the one to fix it.
- **The opposite symptom**, which reads nothing like an encoder problem and is not one: after dark the sensor slows below the frame rate asked of it. The picture still moves and the bitrate still looks fine while only the frame rate has gone.
- **What you see** and **Checking it**, laid out the way the viewer-limit section next to it is.

Plus an entry in `trouble-majestic.md` — which already carries the reboot symptom this one feeds — and the two bare `minQp`/`maxQp` lines in the config reference given the sentence that matters.

## On framing

Written symptom-first, starting from the tile an operator is looking at when they come asking rather than from the thing that was added. The metric table sits under *Checking it* rather than up front: the numbers are for whoever is scraping them, and the sentence on the page is for everyone else.

## Scope

English only. `ru/` mirrors 31 of 144 pages and none of the Majestic ones, so this matches practice. No `README.md` change — these are sections in existing pages.


widgetii:
Both findings addressed, one of them twice.

**Remote commands rejected** — correct and fixed. Every example addressed to a camera now carries `-u root:PASSWORD`, matching the convention the rest of the wiki already uses; localhost examples stay bare, as they are elsewhere.

**Private rate states** — the finding is right in principle and was fixed before this re-review ran; the excerpt it quotes is the line that was removed. What replaced it named no states and enumerated no values, but it did still imply an ordering, so that has gone too. The row now says only: `0` while the channel is meeting what was asked of it, non-zero while it is not, and the Dashboard and the log for which finding it is.

One part of the recommended fix I have not taken, deliberately: removing the absence rule. That `venc0_rc_state` is **absent** rather than zero when the camera has no opinion is a property of the metric as served, not a classification of anything — and it is the single fact a monitoring system most needs, because reading absent as zero makes every camera too old to publish it look healthy. Dropping it would leave the page quietly wrong in the direction that costs someone an outage. It stays, alongside the advice to alert on presence and non-zero rather than on a value.

For calibration, the neighbouring viewer-limit section documents `live_backlog_pressure` as *"0 while there is room, rising to 4 when the camera is trimming hard"* — the same shape as what is left here.

PR #553, opened 2026-09-19
Backing up your stock firmware: one page, six routes

openipc-ai:
## Problem

There is no "back up your stock firmware" page in this wiki, and nothing anywhere says where to send a dump.

The recipe exists in seven places and nowhere as a page: `en/old-manual.md` has the `sf read` + `tftp` form (filed under *Very old full manual*), `en/install-hisi.md` the `tftpput` variant, `en/install-xm530.md` the serial hexdump, `en/help-uboot.md` the no-TFTP and SD-card routes, `en/faq.md` the NFS one, and two `device-*.md` pages their own. `en/installation.md` Step 6 describes the step in prose and deliberately gives no addresses.

`ipctool backup` / `upload` / `restore` exist and are documented nowhere at all — grepping the whole wiki for them returns zero hits.

That gap shows up in OpenIPC/firmware as cameras with no way back: #2299, #2381, #1908.

## What this adds

`en/backup-stock-firmware.md`, ordered by **what the camera still lets you do** rather than by method, so a reader with a half-dead board finds their row instead of reading six procedures that assume more than they have:

| Situation | Route |
| --- | --- |
| Stock firmware runs, you have a shell | A — `cat /dev/mtdblock*`, NFS |
| `ipctool` is on it | B — `ipctool backup` / `upload` / `restore` |
| Bootloader console, network works | C — U-Boot over TFTP |
| Bootloader console, no network | D — to an SD card, or E — over the serial line |
| No console, or already dead | F — CH341A and a clip |

Every command is taken from the existing pages, not invented; each route links back to the page it came from.

Two things written down here for the first time:

- **`ipctool backup` / `upload` / `restore`.**
- **The three-argument `tftp` trap.** On HiSilicon's combined `tftp`, the three-argument form is an *upload* — which is exactly what the backup step wants, and exactly wrong when the habit carries over to flashing, where U-Boot sends RAM to the server and writes nothing to the camera. That is how #2299 destroyed a flash while believing it had installed an image.

Plus a "check the dump before you trust it" section: exact size, a compression sanity check, and `strings dump.bin | grep ^ethaddr`.

## Why now

OpenIPC/firmware is getting issue forms that ask for a vendor dump, including a required one on the new-board form. They need somewhere to point. The last section covers the attachment rules those forms impose — compress it, `.bin` is not accepted, 25 MB cap, link a 128 MB NAND image instead.

## Checks

- All relative links and anchors resolve, including `help-uboot.md#prepare-the-environment`.
- All six in-page anchors in the routing table resolve.
- Registered in `README.md` under **Installation**, before *Old installation guides*.
- `cspell.yaml` ignores fenced and inline code, so the command blocks need no dictionary additions.


openipc-ai:
All seven addressed in f936a91 — every one of them was real, and five would have cost someone a camera or a credential. Thank you.

**2 · `cat /dev/mtdblock*` scrambles ≥10 partitions.** Correct, and the note underneath made it worse by recommending that glob for exactly that case. Now reads the order from `/proc/mtd`:

```bash
for m in $(sed -n 's/^mtd\([0-9]*\):.*/\1/p' /proc/mtd); do cat /dev/mtdblock$m >> $out; done
```

**3 · NFS mount not checked.** Correct, and the consequence is the bad one: a failed mount leaves an ordinary directory *on the camera*, `mv` succeeds into it, and the only copy of the stock firmware is erased by the reflash it was taken for. Guarded on both the mount and `/proc/mounts`, with that spelled out.

**4 · `tftpput` hard-coded `0x42000000` and 8 MB.** Correct — a data abort on the 0x80000000 family and half a backup on 16 MB. I had lifted it verbatim from `install-hisi.md` without harmonising it. Now uses `${baseaddr}`/`${flashsize}` with the reason stated.

**5 · SD route wrote `0x4000` regardless of size.** Correct. I condensed `help-uboot.md`'s separate 8 MB and 16 MB blocks into one and lost the variance, while my own `dd` line used `count=32768`. Now spelled out separately across all three steps.

**6 · NAND cameras routed at `sf`.** The best finding here, and a flaw in the premise rather than a slip: I routed by *access* and never asked which flash. Added a "which flash?" preamble, marked C–E as SPI NOR-only, added **route G** — boot something with a shell and use `nanddump -f ... /dev/mtd0`, which this wiki already documents in `fpv-sigmastar.md` — and gave route F the matching caveat that plenty of cheap programmer software cannot do SPI NAND at all.

**1 · Real MAC.** Correct under rule 2. It came from ipctool's README, but that is not a reason to repeat it. Replaced with a locally administered `02:00:00:00:00:01`.

**7 · Publishing a dump publishes credentials.** Correct, and it reached further than this page. A full image carries the root password hash, usually the Wi-Fi SSID and passphrase the vendor firmware last joined, the device key, the serial and the MAC — and this wiki's own `research-security.md` lists the tools that crack such hashes, so "only a hash" is not protection.

The page now says what is in a dump, says what to change first, and makes *available on request* a respectable answer with a private handover behind it.

It also changed the companion PR. **OpenIPC/firmware#2454's new-hardware form was requiring that upload** — so a reporter either published their own Wi-Fi password or could not file. Fixed there too: the ask stays firm via a required dropdown recording how the dump will be provided (attached / on request / linked / cannot produce one), and the upload itself is optional. The answer is always known; nobody is forced to publish.

Verified: all in-page anchors and file links resolve.

PR #554, opened 2026-09-19
Frigate: let the camera publish, and turn off go2rtc's STUN

widgetii:
## Why

The Frigate page has only ever described Frigate opening connections **to** the camera. That is the right default, and the wrong one on a small board.

A camera holds unsent video for anything reading it over TCP, and that is the memory it runs out of first. Measured on a 27 MB gk7205v200 serving Frigate two streams:

| how Frigate gets the video | connections to the camera | of its live-client budget |
|---|---|---|
| RTSP, pulled (what the page describes today) | 2 | **2.5 MB** — about half |
| WHIP, pushed | **0** | **none** |

A WebRTC publish is UDP, so there is no send buffer for a slow link to fill and nothing for a stalled reader to grow. The page says what it costs too — about 1.4 MB more memory spent publishing than serving. It is memory that cannot be grown from the far end, which is the whole point.

## The part nobody would guess

go2rtc will not answer a WHIP offer until it has finished gathering its own ICE candidates, and it ships pointed at two public STUN servers. Publishing to a machine on your own LAN therefore waits for two internet round trips whose answer is the camera as seen from outside — which is not how anything on that LAN reaches it.

Offer to answer, measured against go2rtc 1.9.10 with an independent client on loopback:

| go2rtc `webrtc` config | time to answer |
|---|---|
| as shipped | ~11 s |
| `candidates` set | ~5 s |
| **`ice_servers: []`** | **~0.002 s** |

So the page gives that line the weight it deserves rather than listing it as a tuning option. Without it the camera spends ten seconds of every start-up waiting, and a camera with no route to the internet waits the full timeout on every session, forever.

## What it does not do

It does not remove go2rtc, and the page says so plainly. Frigate's detect and record are ffmpeg processes that read RTSP; no camera speaking WebRTC changes that. go2rtc is what turns the pushed stream back into something they can read — only the direction of the video changes.

It is also not a shape Frigate's own docs cover, so the page frames it as the option it is: worth it on a board short of memory or a link where viewers stall, not worth the deviation otherwise.

## Also

A pointer from the outgoing-destinations section of the streamer page, since a slow answer looks the same from any WHIP endpoint, and the camera now names which of the two failures it saw rather than reporting both as unreachable.

Verified end to end on the hardware the numbers came from: camera publishing both channels into the go2rtc inside Frigate, Frigate consuming them onward into detect and record, two to three seconds from the encoder starting to both channels live.


widgetii:
Two of three taken; the third partly.

**Verbatim log strings** — right, and it was a rule I had already applied on another page and then broke on this one. A closed component's log lines are exactly what the disclosure gate keeps out, and a merged page keeps them forever. The note now says what the camera *distinguishes* — an endpoint never reached against one reached and silent — and where that shows, without printing either line.

**The 1.4 MB figure** — right, and worse than the finding knew. It came out of two `MemAvailable` readings taken from runs that had not started from the same state, which is not a measurement and should not have been written as one. Dropped. What replaces it is the number that is actually reproducible and now carries its method: `live_backlog_reserved_bytes` off `/metrics`, **2621440** with two TCP readers and **0** with a push, alongside the board, its RAM, both stream sizes and which role reads which. The qualitative trade is still stated because it is true and it matters — a publishing session does spend memory of its own — but the claim is now about *who can spend* the camera's memory rather than a total nobody can check.

**The build requirement** — the premise does not hold for the builds this page is written for. Lite carries WebRTC; that is what the cameras in the examples run, and what the measurements were taken on. So I have not restructured the page around a warning that would be wrong for nearly every reader of it. But a build without it is possible, so one sentence now says so at the point the camera is configured, in the form the rest of the wiki already uses for this: an absent key means an unsupported build, and the API refuses the address rather than accepting a setting nothing would act on.

PR #558, opened 2026-09-20
Day/night: the two park switches became one mode each

openipc-ai:
Two `nightMode` keys documented here do not exist any more, and three that do
exist are documented nowhere.

The daemon folded each actuator's park boolean into a single three-state mode —
four combinations covered three states — so `nightMode.irCutEnabled` and
`nightMode.backlightEnabled` were replaced by `nightMode.irCut` and
`nightMode.backlight`, each `off`, `manual` or `auto` and defaulting to `auto`.
The wiki never caught up, so a reader following it was writing keys the camera
does not have.

Asked of a current camera (hi3516av300):

```
nightMode.irCutEnabled        404
nightMode.backlightEnabled    404
nightMode.irCut               200  auto
nightMode.backlight           200  auto
nightMode.transitionDelayMs   200  0
```

### What changed

**`en/majestic-streamer.md`** — the *"Two park switches"* paragraph becomes a
section, *Who moves the filter and the lamp*, with the three values in a table.
It also says what the old paragraph could not: **none of the three means
"nothing is wired here"**. All of them keep the pin numbers configured, and what
they choose is who decides — worth knowing before reading a pink daylight
picture as a wiring fault rather than as a parked filter.

The same section now covers `nightMode.transitionDelayMs` (default `0`, range
0-2000), which was documented nowhere: the pause inside a switch, for the frame
or two that can arrive in colour while the filter is still swinging.

**`en/majestic-config.md`** — the example `nightMode` block loses the two dead
keys, gains `irCut`/`backlight` with the three values explained, and gains the
transition delay beside the other switching settings.

Nothing here changes behaviour for a camera that was never told otherwise: the
default is what every camera already did.

Spell-checked against the repo's own `cspell.yaml` — clean.


openipc-ai:
Finding taken, and fixed in b226fce.

The reviewer is right that "150 was enough on the one camera that wanted it" is
an anecdote wearing a measurement's clothes — no board, no sensor, no method,
nothing a reader could reproduce. I cannot supply the measurement either: the
camera I verified this PR against has nothing wired to its filter, so there is
no swing here to time.

So the number is gone rather than dressed up. How long a filter takes to swing
is a property of the board, and the page now says so instead of offering a
starting point: leave it at `0`, raise it only if a coloured frame actually
shows as day returns, and then in small steps. The commented example still
carries a value, as every other commented key in that block does, but its note
reads *board-specific* rather than implying a recommendation.

The rest of the PR rests on things anyone can check from outside a camera — the
keys and their defaults and ranges, and the two manual endpoints, all asked of a
running camera before being written down.

